PR #13, opened 2020-07-10
Fix Wifi AP issues

sbhklr:
Due to a bug in the wifi AP setup function  the netmask wasn't set correctly and some users couldn't connect to the Portenta. This PR fixes it and also makes the network parameters configurable.


sbhklr:
Ciao @facchinm, I just tested it with Nacho, this PR solves the connection issues.

PR #14, opened 2020-07-10
WiFi UDP & Multicast Functionality

sbhklr:
This PR implements some missing functions for WiFiUDP functionality. Previously it didn't compile because not all virtual functions were implemented. Some testing still need to be done.

### DIFF
--- a/libraries/WiFi/src/WiFiUdp.cpp
+++ b/libraries/WiFi/src/WiFiUdp.cpp
@@ -67,11 +67,14 @@ int arduino::WiFiUDP::parsePacket() {
     if (ret == NSAPI_ERROR_WOULD_BLOCK) {
         // no data
         return 0;
+    } else if(ret == NSAPI_ERROR_NO_SOCKET){
+        // socket was not created correctly.
+        return -1;
     }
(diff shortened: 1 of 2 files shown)


jwestmoreland:
I built with this - the wifi-udp branch - it apparently doesn't work - or maybe it doesn't work with multicast (yet) since my sketch is using multicast.  Note, if I use this branch, my sketch will reset every few minutes or so.  If I just include the WiFiUDP files that are included with the 1.2.1 version in this repo - the sketch doesn't reset but still no UDP via multicast.

Just an FYI. 

I'm not sure the debugger works either with the portenta_m7 yet - but I will look into the configs on that to see if there is one yet that supports the portenta.

Regards,
John W.

sbhklr:
@facchinm Looks like we can merge this and close https://github.com/arduino/ArduinoCore-mbed/issues/15.

jwestmoreland:
Sebastian,

Maybe change title to:  Wifi UDP & Multicast Functionality ?

Thanks,
John

jwestmoreland:
Thanks Martino and Sebastian!

PR #18, opened 2020-07-14
Portenta: WiFi: remove debug on Serial1

facchinm:
Logs can be printed by calling
```
whd_print_logbuffer();
```

To change default port from `Serial1` to another transport (`Serial` for example), simply declare
```
REDIRECT_STDOUT_TO(Serial);
```
at the top of the sketch.


jwestmoreland:
Thanks again for this Martino!  Came in handy when going back to a previous rev to make sure WiFi was OK,
John

PR #20, opened 2020-07-15
Add CI

umbynos:
I've added a github CI workflow useful to compile the examples from the libraries that come with the core.
I did some try and error investigation regarding the libraries in the core and the compatibility with nano ble or portenta boards: some libraries are compatible with both the portenta platform and the nano33ble platform, some others are not. Thus the compile process is done only for the compatible examples.


umbynos:
Added also the report deltas action: It should report the deltas in an open PR

PR #23, opened 2020-07-22
Add easy to remember alias for LL_RCC_ForceCM4Boot

sbhklr:
I've added an alias for force-booting the M4 which is less cryptic than LL_RCC_ForceCM4Boot.
Not sure if Arduino.h is the best place to put it. I'm open to feedback. @facchinm


facchinm:
This should be moved to PORTENTA_M7 `variant.h` since neither PORTENTA_M4 nor Nano33BLE have the chance to call that command :wink:

PR #63, opened 2020-10-02
Fix UART::write(const uint8_t *, ...) return size

manchoz:
The UART::write(const uint8_t* c, size_t len) method now returns the actual number of written bytes.

Tested in production on Project Challenge Board (Arduino Edge Control).


sbhklr (approved):
LGTM

facchinm (approved):

PR #71, opened 2020-10-15
Nudge user to use the M7 for bootloader and WiFi updates

sbhklr:
In my opinion it doesn't make much sense to allow the users to upload the bootloader / wifi FW update sketch to the M4 as they would need to force-boot the M4 on the M7 just for that. If someone accidentally chooses the M4 core they may wonder why nothing happens.
This PR forces the users to use the M7 for those updates.


facchinm (approved):

sbhklr:
@per1234 What's your suggestion on making the CI happy with this?

per1234 (approved):
Sorry for the slow response. The CI configuration change looks good.

sbhklr:
@per1234 I removed the the system sketches from the M4 compile workflow. Are we good with those changes?

PR #77, opened 2020-10-26
PluggableUSBMSD: fix functionality on Portenta

facchinm:
External flash can be read from a host PC with a sketch resembling this one

```c++
#include "PluggableUSBMSD.h"
#include "QSPIFBlockDevice.h"
#include "FATFileSystem.h"

static QSPIFBlockDevice bd(PD_11, PD_12, PF_7, PD_13,  PF_10, PG_6, QSPIF_POLARITY_MODE_1, 40000000);

void USBMSD::begin()
{
    int err = getFileSystem().mount(&bd);
    if (err) {
        err = getFileSystem().reformat(&bd);
    }
}

mbed::FATFileSystem& USBMSD::getFileSystem()
{
  static mbed::FATFileSystem fs("fs");
  return fs;
}

USBMSD MassStorage(&bd);

void setup() {
  MassStorage.begin();
}

void loop() {
}
```

### DIFF
--- a/libraries/USBMSD/PluggableUSBMSD.h
+++ b/libraries/USBMSD/PluggableUSBMSD.h
@@ -22,6 +22,8 @@
 #include "USBDescriptor.h"
 #include "USBDevice_Types.h"
 #include "platform/Callback.h"
+#include "drivers/internal/PolledQueue.h"
+#include "drivers/internal/Task.h"
 #include "BlockDevice.h"
 #include "Mutex.h"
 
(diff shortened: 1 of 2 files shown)


sbhklr:
@facchinm This is very cool! It works beautifully with the internal flash. I tried getting it to work with the SD card without success:
```
#include "SDMMCBlockDevice.h"
SDMMCBlockDevice bd;
```
It gets stuck at the `mount` function. Is this supported?

facchinm:
On SDCard there's a bug lurking somewhere; I couldn't make it run either but didn't spend much time since the goal was exposing the QSPI flash :wink:

Issue #78, opened 2020-10-26
WiFi AP status doesn't change when a client connects


sebromero:
In the example code there is this snipped which checks for new clients connecting to the AP:
```
// compare the previous status to the current status
  if (status != WiFi.status()) {
    // it has changed update the variable
    status = WiFi.status();

    if (status == WL_AP_CONNECTED) {
      // a device has connected to the AP
      Serial.println("Device connected to AP");
    } else {
      // a device has disconnected from the AP, and we are back in listening mode
      Serial.println("Device disconnected from AP");
    }
  }
```
The status doesn't seem to change even when a client successfully connects. I couldn't find any references in the WiFi library where `WL_AP_CONNECTED` is set. The example worked with an old version of the core as there was a hardcoded value being returned by the status function:
```
uint8_t arduino::WiFiClass::status() {
    // @todo: fix
    return WL_CONNECTED;
}
```
Probably we need to look into `register_event_handler` function of the `WhdSoftAPInterface` class. The documentation is quite limited and I couldn't find an example of how to implement it yet. https://os.mbed.com/docs/mbed-os/v6.3/feature-i2c-doxy/class_whd_soft_a_p_interface.html#a2ab8fabaa3c42aa60fc3d4438368d43f

PR #79, opened 2020-10-26
Implement event handler for AP connections

sebromero:
Tries to solve https://github.com/arduino/ArduinoCore-mbed/issues/78
```
if(event_header->event_type == 8){ //8 = connect
    WiFi._currentNetworkStatus = WL_AP_CONNECTED;
    return nullptr;
}

if(event_header->event_type == 12){ //12 = disconnect
    WiFi._currentNetworkStatus = WL_AP_LISTENING;
    return nullptr;
} 
```
@facchinm I couldn't figure out where those events for connect/disconnect are defined so I went with the raw values. Any hints?

### DIFF
--- a/libraries/WiFi/src/WiFi.cpp
+++ b/libraries/WiFi/src/WiFi.cpp
@@ -71,8 +71,10 @@ int arduino::WiFiClass::beginAP(const char* ssid, const char *passphrase, uint8_
         }                
 
         // Default Event Handler
+        whd_driver_t whd_driver = ifp->whd_driver;
+        WHD_IOCTL_LOG_ADD_EVENT(whd_driver, event_header->event_type, event_header->flags, event_header->reason);
+        
         if ((event_header->event_type == (whd_event_num_t)WLC_E_LINK) || (event_header->event_type == WLC_E_IF)) {
-            whd_driver_t whd_driver = ifp->whd_driver;
             if (osSemaphoreGetCount(whd_driver->ap_info.whd_wifi_sleep_flag) < 1) {
                 osStatus_t result = osSemaphoreRelease(whd_driver->ap_info.whd_wifi_sleep_flag);
                 if (result != osOK) {
--- a/libraries/WiFi/src/WiFi.cpp
+++ b/libraries/WiFi/src/WiFi.cpp
@@ -44,34 +44,49 @@ int arduino::WiFiClass::begin(const char* ssid, const char *passphrase) {
 
 int arduino::WiFiClass::beginAP(const char* ssid, const char *passphrase, uint8_t channel) {
 
-#if defined(ARDUINO_PORTENTA_H7_M7) || defined(ARDUINO_PORTENTA_H7_M4)
-    _softAP = WhdSoftAPInterface::get_default_instance();
-#endif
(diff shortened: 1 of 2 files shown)


facchinm:
@sbhklr they can be retrieved from https://github.com/arduino/ArduinoCore-mbed/blob/master/cores/arduino/mbed/targets/TARGET_STM/TARGET_STM32H7/TARGET_STM32H747xI/TARGET_PORTENTA_H7/COMPONENT_WHD/inc/whd_events.h#L42 .
As always, outstanding stuff! My only concern is that we are overriding default handler ([whd_default_handle_softap_events](https://github.com/arduino/mbed-os/blob/portenta-mbed-6.2.0/features/netsocket/emac-drivers/TARGET_STM/COMPONENT_WHD/interface/WhdSoftAPInterface.cpp#L37)) which takes care of sleeping... Shouldn't be a huge problem, we can probably deal with it later.

sbhklr:
There doesn't seem to be a way to call the default handler beforehand. The function is private and there is no API for retrieving any registered event handler (couldn't find one) which could be used to anonymously call the default handler.
As the handler is just a few lines of code I copied into the `beginAP` function. Not so elegant but probably better than not dealing with the sleep events.

facchinm:
@sbhklr copying the content of the default handler is fine, but since the function grew significantly I'd replace the lambda with a proper function pointer and extract it. What do you think?

sbhklr:
@facchinm Fair enough, done.

rchard-84:
I've run into this issue as well trying to move code from an UNO over to an H7.  When, very roughly, might this fix show up in a release.  Thanks for finding and fixing the issue.

sebromero:
I think we're ready to merge this one.
@facchinm any time frame for the next core release?

sebromero:
@rchard-84 Have you tested this PR? It works on my end, but always good to have more testers :-)

facchinm:
> I think we're ready to merge this one.
> @facchinm any time frame for the next core release?

Ok for merging :slightly_smiling_face: Next release is due in a week more or less (after https://github.com/ARMmbed/mbed-os/pull/13826 gets merged)

rchard-84:
I did not test the PR.  I lack the knowhow to go from your fix to a
compiled library that I can test on the portenta. Is there a walk through
someplace?  I'd be happy to try stuff out.

sebromero:
@rchard-84 Basically you can check out this repository and put it in your sketchbook. Then you can select that version of the core from the menu in the IDE.
image

There might be some documentation on that. @per1234 if anyone, then you'll know it :-) Any hints?

per1234:
As for the technique of using a copy of the core in the sketchbook for development/beta testing, there is an interesting discussion about it here:
https://github.com/arduino/Arduino/issues/9237

As for official documentation of the manual core installation process, there isn't any that I'm aware of beyond a quick mention in the Arduino Platform Specification (the preferred term for this is "platform"):
https://arduino.github.io/arduino-cli/latest/platform-specification/
>Platforms add support for new boards to the Arduino development software. They are installable either via Boards Manager or manual installation to the hardware folder of Arduino's sketchbook folder (AKA "user directory").

I would expect to find manual platform installation documented here: https://www.arduino.cc/en/Guide/Cores, but it only talks about Boards Manager installation.

sebromero:
Thanks @per1234 ! I've asked content to add documentation to that page.

PR #80, opened 2020-10-27
Makes available() non-blocking (fixes #76)

### LINKED ISSUE #76: WiFiClient client = server.available(); blocking
is this the preferred behavior?
if it is not:
add in WiFiServer.cpp above line 38:
sock->set_blocking(false);

Best Regards
Andreas D.

sebromero:
This PR makes `available` non-blocking and thus fixes https://github.com/arduino/ArduinoCore-mbed/issues/76. I'm not sure if we should also go with a client socket re-use approach as in the NINA library: https://github.com/arduino-libraries/WiFiNINA/blob/master/src/WiFiServer.cpp#L48
I currently struggle to see the use case for this as it's the caller's responsibility to keep the client instance which wraps the socket. Then again my networking knowledge has faded, maybe I'm overlooking something :-)

Btw. using the `WiFiClient` class on a server side application is a bit confusing imho

### DIFF
--- a/libraries/WiFi/examples/PortentaWiFiFirmwareUpdater/PortentaWiFiFirmwareUpdater.ino
+++ b/libraries/WiFi/examples/PortentaWiFiFirmwareUpdater/PortentaWiFiFirmwareUpdater.ino
@@ -36,7 +36,7 @@ void setup() {
     Serial.println("No filesystem containing the WiFi firmware was found.");
     Serial.println("Usually that means that the WiFi firmware has not been installed yet"
                   " or was overwritten with another firmware.\n");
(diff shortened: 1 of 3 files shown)


sidprice:
I am working with Portenta H7 and using WiFiUdp class and the blocking of "available" and "parsePacket" is an issue, in that the hardcoded socket timeout of 1000mS means in the absence of network data being received the "Loop" function is only run once per-second. It seems, the client should have control over whether the socket is blocking or not. This could be implemented by exposing a public method allowing the client to set the socket timeout since a value of 0mS would make it non-blocking.

PR #99, opened 2020-12-11
UART: Allow baud rate changes

marcusbirkin:
Allows begin() to alter the baud rate if _serial objected already created.

### DIFF
--- a/cores/arduino/Serial.cpp
+++ b/cores/arduino/Serial.cpp
@@ -75,6 +75,8 @@ void UART::begin(unsigned long baudrate, uint16_t config) {
 void UART::begin(unsigned long baudrate) {
 	if (_serial == NULL) {
 		_serial = new mbed::UnbufferedSerial(tx, rx, baudrate);
+	} else {
+		_serial->baud(baudrate);
 	}
 	if (rts != NC) {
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @marcusbirkin , thanks for the contribution!
The usual pattern to change baud rate in the arduino world is
```
Serial.begin(baud);
.
.
.
Serial.end();
Serial.begin(new_baud);
```
that would still work with the actual codebase. 

Are you aware of any library/project that don't call `end()` ? In that case I'd merge the PR because it's just fine; otherwise I'd rather keep the actual behaviour, even if the failure is silent.

marcusbirkin:
`end()` seems quite wasteful and slow, it deletes the `_serial` object, only for it to be reallocated and initialised again with a `begin()`.
The AVR HAL (https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/HardwareSerial.cpp) allows for baud rate change via `begin()`.

facchinm:
It is indeed wasteful and slow, but how many times are you expecting to change the baud rate? Probably changing twice in the lifetime of the application is already a lot.
Anyway, I'm merging the patch for the principle of least astonishment :slightly_smiling_face: Thanks for the contribution!

Issue #96, opened 2020-12-09
Thread example RPC_m4


JSCLadon:
Hi there,
since I am learning, I really rely on the examples.  The following example seems to be not working correctly: 
https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/RPC/examples/RPC_m4/RPC_m4.ino

It produces the following output: 

17:30:10.131 -> Calling subtract -33
17:30:10.877 -> Calling subtract -33
17:30:11.456 -> Calling add Calling subtract 57
17:30:12.507 -> -33
17:30:13.188 -> Calling subtract -33
17:30:13.937 -> Calling subtract -33
17:30:14.475 -> Calling add Calling subtract 57
17:30:15.553 -> -33
17:30:16.228 -> Calling subtract -33
17:30:16.978 -> Calling subtract -33
17:30:17.555 -> Calling add Calling subtract 57
17:30:18.605 -> -33
17:30:19.282 -> Calling subtract -33
17:30:20.023 -> Calling subtract -33
17:30:20.596 -> Calling add Calling subtract 57
17:30:21.646 -> -33
17:30:22.352 -> Calling subtract -33
17:30:23.101 -> Calling subtract -33
17:30:23.641 -> Calling add Calling subtract 57
17:30:24.693 -> -33
17:30:25.373 -> Calling subtract -33
17:30:26.148 -> Calling subtract -33
17:30:26.690 -> Calling add Calling subtract 57
17:30:27.738 -> -33
17:30:28.447 -> Calling subtract -33
17:30:29.190 -> Calling subtract -33
17:30:29.731 -> Calling add Calling subtract 57
17:30:30.784 -> -33

I guess, that is not, what it is supposed to do. M7 is never really calling the functions. Is that correct?

facchinm:
Hi @JSCLadon ,
the example was never meant to be final, sorry.
Two things here:
* I'm attaching the same example but better commented :slightly_smiling_face: 
* The `printf` called by CM7 in the original code was printing to the internal UART, not on the USB, so it was "invisible" unless you connected a standalone USB-to-serial adapter

Here's the new code
```c++
#include "Arduino.h"
#include "RPC_internal.h"

String currentCPU() {
  if (HAL_GetCurrentCPUID() == CM7_CPUID) {
    return "CM7: ";
  } else {
    return "CM4: ";
  }
}

int add(int a, int b) {
  Serial.println(currentCPU() + "executing add()");
  delay(1000);
[...]

PR #103, opened 2020-12-15
Make RPC example easier to understand

sebromero:
"Fixes" https://github.com/arduino/ArduinoCore-mbed/issues/96

### DIFF
--- a/libraries/RPC/examples/RPC_m4/RPC_m4.ino
+++ b/libraries/RPC/examples/RPC_m4/RPC_m4.ino
@@ -1,58 +1,110 @@
 #include "Arduino.h"
 #include "RPC_internal.h"
 
-#ifdef CORE_CM7
-int add(int a, int b) {
-  printf("calling add on M7\n");
-  delay(1000);
+using namespace rtos;
+
+Thread subtractThread;
+
+/**
+ * Returns the CPU that's currently running the sketch (M7 or M4)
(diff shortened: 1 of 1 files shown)


JSCLadon:
I am trying to pass a class Value1P from M4 to M7. 
The calass is defined as: 
`class Value1P { 
    int lineID;
    double current;
    double tension;
    Value1P() {
        lineID = 0001;
        current = 0;
        tension = 0;
    };
    ~Value1P() {};
};`
I adjusted the example and got the error message: " Value1P has no member named 'msgpack_unpack' v.msgpack_unpack(o.convert());"

I got stuck. If you improve your examples, could you consider an example, were not only a plain int is passed from one core to another?

That would be just wonderful!

Thanks in advance!

facchinm:
Hi @JSCLadon ,
values being passed from one function to another must be serializable in msgpack (see https://github.com/msgpack/msgpack/blob/master/spec.md#formats for the supported ones). The solution is usually a packer/unpacker from basic types to the actual object before passing it to the function.
@sebromero do you think that adding an example like that would make sense?

sebromero:
@facchinm Yes, I think it would be useful. I would probably put it into a separate sketch but I expect other people to have the same issue.

JSCLadon:
This is a perfect example! Thanks a lot!

facchinm (approved):

PR #109, opened 2020-12-21
Wifi example

don:
Modified version of WiFiWebClient from WiFiNINA

### DIFF
--- a/libraries/WiFi/examples/WiFiWebClient/WiFiWebClient.ino
+++ b/libraries/WiFi/examples/WiFiWebClient/WiFiWebClient.ino
@@ -1,18 +1,13 @@
-
 /*
   Web client
 
- This sketch connects to a website (http://www.google.com)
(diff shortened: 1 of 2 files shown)


libraries/WiFi/examples/WiFiWebClient/WiFiWebClient.ino
```diff
@@ -0,0 +1,108 @@
...
+
+ This example is written for a network using WPA encryption. For
+ WEP or WPA, change the Wifi.begin() call accordingly.
+
+ Circuit:
+ * Arduino Portenta H7
+
+ created 13 July 2010
+ by dlf (Metodo2 srl)
+ modified 31 May 2012
```
sebromero, line 14: @don If you made changes to it, you may want to put your name here. Looks good otherwise.

sebromero (changes requested):

PR #120, opened 2021-01-13
Changed repository clone path to avoid shadowing

pennam:
@facchinm i've only updated repository clone path to avoid shadowing issue


pennam:
closing and opening another one with correct source branch

PR #140, opened 2021-02-17
include Arduino.h in Portenta_Camera/camera.h

rfreedman:
Add include of Arduino.h in Portenta_Camera/camera.h to allow inclusion of camera.h in a .cpp file without having to explicitly include Arduino.h in the .cpp file


facchinm:
LGTM! Thanks, merging

PR #143, opened 2021-02-22
Fix buffer sync issue in camera raw bytes example

sebromero:
Adds a very basic synchronization to the Arduino and Processing sketch to make sure the buffer doesn't get out of sync.
Here is the before / after comparison:

Before

After

### DIFF
--- a/libraries/Portenta_Camera/examples/CameraCaptureRawBytes/CameraCaptureRawBytes.ino
+++ b/libraries/Portenta_Camera/examples/CameraCaptureRawBytes/CameraCaptureRawBytes.ino
@@ -4,8 +4,7 @@ CameraClass cam;
 uint8_t fb[320*240];
 
 void setup() {
-
-  Serial.begin(921600);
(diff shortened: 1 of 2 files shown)


sebromero:
@iabdalkader Wdyt?

iabdalkader:
You might want to add a timeout in the pde.

sebromero:
> You might want to add a timeout in the pde.

@iabdalkader Can you elaborate? Timeout for what condition?

iabdalkader:
@sebromero If the sketch misses even a single byte, the cam and the sketch will deadlock. Adding a 1 second ? timeout in the pde may allow it to recover from that without restarting the sketch. Just a suggestion but otherwise it looks good.

sebromero:
@iabdalkader Makes sense. Added it 👍

facchinm (approved):

gargoyle777:
i'm still having the exact same problem. i'm using the last released exemple by the way.

PR #146, opened 2021-02-23
Add Nano33_updateBootloader sketch to support SoftDevices

giulcioffi:
This sketch allows to update the Bootloader to support SoftDevices and is related to [bcmi-labs/zephyr/PR#4](https://github.com/bcmi-labs/zephyr/pull/4).
The sketch can be uploaded at 0x10000 (as in the standard bootloader and core) and then performs the following operations:
- write the MBR at address 0x0
- write the new bootloader at address 0xE0000
- write the Soft Device at address 0xA0000
- enable the bootloader by writing its address to the UICR memory (0x10001014)
- write a magic (flag for the firmware update) at address 0xFF000 with the info related to SoftDevice (flash address and length)
- reboot
The operations performed in flash when the sketch is uploaded are the following:
image


facchinm (approved):

libraries/Nano33_System/examples/Nano33_updateBootloader/Nano33_updateBootloader.ino
```diff
@@ -0,0 +1,171 @@
...
+
+const unsigned int magic = 0x5f27a93d;
+
+mbed::FlashIAP flash;
+
+void setup() {  
+  Serial.begin(115200);
+  while (!Serial) {}
+  Serial.println("Do you want to update only the SoftDevice? Y/[n]");
+  Serial.println("Yes: upload only the binary of the DoftDevice");
```
sebromero, line 47:
```suggestion
  Serial.println("Yes: upload only the binary of the SoftDevice");
```

sebromero:
@giulcioffi @facchinm Shall we split this up into a bootloader update sketch and a soft device firmware? Might be a bit easier to understand and find for the the users. Also soft device will probably be flashed more often than the bootloader.

giulcioffi:
Hi @sebromero, it may be a good idea. My only concern is that, from the user point of view, it needs to be clear that a soft device firmware update works only if the bootloader update is performed before. For this reason, having everything in one sketch can be simpler.
Another option could be to keep everything in a single sketch and let the sketch detect if the bootloader update is necessary or not, instead of asking it to the user.

sebromero:
@giulcioffi What if we split it up and in the soft device sketch we check for the bootloader version and if it doesn't match we ask the user to run the bootloader sketch? At least for the MicroPython/OpenMV use case where we just want people to update the bootloader it would be nice if they had a standalone BL updater and didn't have to think about the Soft Device.
Here's how we check the bootloader version for Portenta: https://github.com/arduino/ArduinoCore-mbed/blob/bd529a0a676e9b6347722e780bb295c669eacbae/libraries/Portenta_System/examples/PortentaH7_updateBootloader/PortentaH7_updateBootloader.ino#L32

giulcioffi:
Hi @sebromero, I split the sketches as you suggested :) 
In 'Nano33_updateSoftDevice.ino' sketch, two defines (`BOOTLOADER_CRC` and `BOOTLOADER_SIZE`) are used to check if the new bootloader is present. Instead of a version check, I went for this solution because the new bootloader at 0xE0000 could be potentially overwritten by some other code...So checking the CRC is a good way to know that the full bootloader binary is still available in the flash.
Could you please test the new sketches and provide a feedback? Thanks :)

sebromero:
@giulcioffi Splendid! I'll try it out.

sebromero:
Hi @giulcioffi We just tested it and it seems that for bossac to work both the bootloader and the softdevice sketch need to be run. Can you confirm this? My understanding was that the bootloader update will allow us to flash OpenMV using bossac.

facchinm:
@sebromero for openmv in particular you need both since it depends on softdevice. For any other sketch the bootloader is enough but you can't tell the difference at "upload time".

giulcioffi:
I confirm what @facchinm said. The updated bootloader will always look first for a valid sketch at 0x10000 and this is what ensures backwards compatibility with the standard sketches uploaded from the IDE. If no valid sketch is found (because maybe a SoftDevice has been placed in that portion of flash), the bootloader looks for a valid sketch at 0x26000 and this is the one that you upload with bossac from command line, as written in the instructions of 'Nano33_updateSoftDevice.ino'.

In the bootloading phase, the start address is set at 0x10000, which is where the IDE starts writing a sketch. The bossac command that I provided for sketches that use Soft Devices (like for OpenMV) specifies instead an offset of 0x16000, which allows to place the sketch at 0x26000.

I hope that this explained a bit better the process :) let me know if you have any other doubt.

sebromero:
Here: https://github.com/giulcioffi/ArduinoCore-mbed/pull/1 I need to test it but don't have a board with me. Will do asap.

sebromero:
Done. PR is ready https://github.com/giulcioffi/ArduinoCore-mbed/pull/1

giulcioffi:
Thank you @sebromero! :) Would you mind updating/reverting also the instructions at the beginning of the sketch? Thanks!

sebromero:
@giulcioffi Sure thing, let me take a look

sebromero:
@giulcioffi Done :-)

PR #138, opened 2021-02-16
PDM: cleanup and fix

facchinm:
* Borrow hardware configuration from OpenMV port for Portenta Vision shield.

PR #149, opened 2021-03-01
PDMSerialPlotter stereo version

pennam:
If https://github.com/arduino/ArduinoCore-mbed/pull/138 gets merged sampleBuffer needs to be increased to support stereo modes. Furthermore Sketch can be modified to print both left and right signals coming from the mics.

Output from will be something like this:

image


facchinm (approved):

pennam:
Tested also on Nano33BLE @16Khz 1 Channel
image

facchinm (approved):

PR #150, opened 2021-03-02
Fixed hot plug hdmi cable event handling. If cable is not present jus…

MaxPayne86:
…t sit and wait instead of resetting anx7625.

### DIFF
--- a/libraries/LittleVGL/examples/Portenta_lvgl/Portenta_lvgl.ino
+++ b/libraries/LittleVGL/examples/Portenta_lvgl/Portenta_lvgl.ino
@@ -124,32 +124,19 @@ static void gpu_fill(lv_disp_drv_t * disp_drv, lv_color_t * dest_buf, lv_coord_t
   }
 }
 
-
 struct edid recognized_edid;
-mbed::DigitalOut video_on(PK_2);
-mbed::DigitalOut video_rst(PJ_3);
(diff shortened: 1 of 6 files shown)


facchinm:
@MaxPayne86 I tested it on a bunch of USBC devices and it works perfectly :muscle: 
There are a couple of places where the same fix needs to be applied, notably https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/LittleVGL/porting.cpp#L133-L155 and https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/doom/doomgeneric_arduino.cpp#L172-L188 .
Would you mind applying the patch there too?

MaxPayne86:
> @MaxPayne86 I tested it on a bunch of USBC devices and it works perfectly 
> There are a couple of places where the same fix needs to be applied, notably https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/LittleVGL/porting.cpp#L133-L155 and https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/doom/doomgeneric_arduino.cpp#L172-L188 .
> Would you mind applying the patch there too?

Please see commit e98a13618eb98c28e7c62999d5bb7c8fbb9ae597

facchinm (approved):

PR #153, opened 2021-03-03
Add A7 pin identifier

sebromero:
Adds the missing A7 pin identifier to be used in Arduino sketches.


sebromero:
The comments for the analog pins in variant.cpp don't match what I see in the pinout diagram.  
E.g. for A5 it says `// A5    ADC1_INP13` but in the data sheet it says `ADC1_INP4`
image
Might be worth double checking.

facchinm:
Rebased and merged as https://github.com/arduino/ArduinoCore-mbed/commit/44d62aeba188a4b08c6547ebdfc34bb944a0bdb9

Issue #155, opened 2021-03-05
Compiler warnings in Wire.cpp and USBCDC.cpp


RMLEARNEY:
I'm sure everyone is seeing these and they're trivial, but thought best to get them noted down:

```
/Users/xxxx/Library/Arduino15/packages/arduino/hardware/mbed/1.3.2/libraries/Wire/Wire.cpp: In member function 'void arduino::MbedI2C::receiveThd()':
/Users/xxxx/Library/Arduino15/packages/arduino/hardware/mbed/1.3.2/libraries/Wire/Wire.cpp:145:12: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      if (c <= sizeof(buf)) {
          ~~^~~~~~~~~~
```

```
/Users/xxxx/Library/Arduino15/packages/arduino/hardware/mbed/1.3.2/cores/arduino/USB/USBCDC.cpp: In member function 'virtual const uint8_t* arduino::USBCDC::configuration_desc(uint8_t)':
/Users/xxxx/Library/Arduino15/packages/arduino/hardware/mbed/1.3.2/cores/arduino/USB/USBCDC.cpp:543:35: warning: narrowing conversion of '((((arduino::USBCDC*)this)->arduino::USBCDC::extraDescriptor != 0) ? 5 : 0)' from 'int' to 'uint8_t {aka unsigned char}' inside { } [-Wnarrowing]
         (extraDescriptor != NULL) ? 0x5 : 0x0, // iInterface
         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
```

zfields:
Resolved by #156

PR #156, opened 2021-03-08
fix: address warnings

zfields:
Addresses type warnings in `Wire.cpp` and `USBCDC.cpp`


zfields:
Resolves #155

facchinm:
Merged, thanks!

PR #157, opened 2021-03-09
Allow for higher resolution with micros()

sebromero:
Makes micros() use a higher resolution timer while keep using a low power timer for millis().
Test sketch:
```cpp
unsigned long start, end;

void runTest(){
  for(int i = 1; i <= 64; ++i){
    start = micros();
    delayMicroseconds(i);
    end = micros();
    Serial.print("Delay: " + String(i) + " measured: ");
    Serial.println(end - start); 
  }
  Serial.println("-----------------------");
}

void setup() {
  Serial.begin(115200);  
  while(!Serial);

  Serial.println(millis());
  runTest();
  Serial.println(millis());
}

void loop() {}
```

### DIFF
--- a/cores/arduino/wiring.cpp
+++ b/cores/arduino/wiring.cpp
@@ -26,22 +26,26 @@
 #include "mbed/rtos/rtos.h"
 #include "mbed/platform/mbed_wait_api.h"
 
-#if DEVICE_LPTICKER
-static mbed::LowPowerTimer t;
-#else
-static mbed::Timer t;
-#endif
-
 using namespace std::chrono_literals;
 using namespace std::chrono;
 
+static mbed::Timer timer;
+
+#if DEVICE_LPTICKER
+  static mbed::LowPowerTimer lowPowerTimer;
+#else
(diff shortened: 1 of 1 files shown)


facchinm:
What do you think about going high res for `micros()` and low power for `millis()` instead? It would turn into a much simpler patch with virtually no drawback ( `micros()` would stop counting during sleep but we can safely add an API to forcefully avoid sleeping when you need precise timing, wrapping `sleep_manager_lock_deep_sleep()` in something a bit more arduinish :slightly_smiling_face: )

sebromero:
@facchinm Legit 😄 . I changed it. Is there a signification penalty in using two timers concurrently?
Am I right in the assumption that some devices may set `DEVICE_LPTICKER=0` as they don't support a LP ticker? I added a fallback for that case.
I'd probably wait with adding a function to avoid sleeping until we have a use case. I currently can't think of one where you need micro seconds resolution during sleep.

facchinm (approved):

PR #162, opened 2021-03-12
Add function to supply VBUS with power

sebromero:
Fixes https://github.com/arduino/arduino-pro-content/issues/17 by adding a function for enabling power on VBUS. `supplyPowerOnVBUS()`

I added specifiers for the architecture in the boards.txt file to allow for conditional compilation but I guess there's a reason why we didn't add them in the first place?


facchinm:
Rebased and merged with https://github.com/arduino/ArduinoCore-mbed/commit/7f2f41bdf0cae4d85617de4005bac5dba91a7aec

PR #166, opened 2021-03-15
Added pins to support breakout

Rocketct:


hpssjellis:
I really appreciate this @Rocketct. I notice several duplicates, but it does allow for easier understanding.

example 

line 12 and 53 both declare PA_4 although 53 adds a few more null's

line 20 and 152 both declare PG_7 etc

Rocketct:
Ciao @hpssjellis the duplicate are required to support the breakoutCarrier, this new product allows to route the M7/M4 pins to some phisical pins on the pcb.
To do this we thought to add all possible pins to the variant and refers to it as a base address, represented by the last  pin listed before LEDB and the offset position of the pin on the new section on that list (e.g PA_4 is obtained as LEDB_pox + 5).
so reason of the duplicate is to maintain the compatibility in case some one change the older  pin on that list( that are used for protocols and for other stuff)

hpssjellis:
@Rocketct I replaced my M7 varient.cpp file with your new one and tried getting the GPIO's running on my Breakout board but without any luck. Is there more to it than just this one file? I am trying codes like:  PC_13, PC_15 etc. They compile which is good, but they don't send any power to the pins yet using the following code. Which tested works fine with 0,1,2, etc Digital pins on the Portenta Board:


image


```
 // Testing the GPIO pins

void myPins(int myGPIO, int myDelay){  
  digitalWrite(myGPIO, HIGH); //external LED on
  delay(myDelay);   
  digitalWrite(myGPIO, LOW); 
  delay(myDelay);
}

void setup() { 
   pinMode(PE_3, OUTPUT);  
   pinMode(LEDB, OUTPUT);
   digitalWrite(LEDB, LOW); //Portenta LED on
   delay(2000);
   digitalWrite(LEDB, HIGH); 
   delay(2000);
}

void loop() {  
   myPins(PE_3, 300);   
   digitalWrite(LEDB, LOW); //Portenta LED on
   delay(1000);
   digitalWrite(LEDB, HIGH); 
   delay(1000);
   
}
```

Rocketct:
yes @hpssjellis you miss another header file, let me prepare everything and I'll link here asap

facchinm:
Rebased and merged as https://github.com/arduino/ArduinoCore-mbed/commit/285b435e57ded7937cca528d8a640cace3c142be

PR #167, opened 2021-03-18
Add menu to split Portenta flash

facchinm:
This patchset allows splitting Portenta flash into different "partitions".
Unfortunately there's no method to keep M7 and M4 options synchronised, so it's up to the user not to flash binaries configured for different partitioning schemes.

The most interesting one is probably 100/0 , where the M4 firmware resides somewhere (not in internal flash) and must be copied to external RAM, address 0x60000000, before calling `bootM4()`

@janjongboom @iabdalkader @pnndra


hpssjellis:
Any instructions on how to test this out @facchinm ?

I just want to run Blinky on the M4 and all the rest of the memory to the M7 to run an ML program and then run Blinky on the M7 and all the rest of the memory to the M4.

iabdalkader:
> I just want to run Blinky on the M4 and all the rest of the memory to the M7 to run an ML program

I have very basic support for that in OpenMV firmware, it requires building the fw from source though, if you want to test it see: https://forums.openmv.io/t/portenta-micropython-on-m7-core-c-on-m4-core/6201/2?u=iabdalkader

hpssjellis:
> > I just want to run Blinky on the M4 and all the rest of the memory to the M7 to run an ML program
> 
> I have very basic support for that in OpenMV firmware, it requires building the fw from source though, if you want to test it see: https://forums.openmv.io/t/portenta-micropython-on-m7-core-c-on-m4-core/6201/2?u=iabdalkader

Thanks @iabdalkader , the instructions seem clear. I was however expecting C++ support for the old Arduino IDE, but from your post it looks like even when there is C++ support it will most likely involve building the C++ from source. Not against building from source, it just adds an extra level of complexity that I was hoping could be configured using an IDE. I think I will wait a bit and see what happens with this.

Anyone from Arduino care to comment?

facchinm:
Hi Jeremy,
I rebased the patches over latest master and added a proper example (under RPC library).
To test this branch, the easiest way is to download the github-provided zip via https://github.com/facchinm/ArduinoCore-mbed/archive/refs/heads/portenta_flash_split.zip, and replace the files from mbed core 2.4.1 (the deprecated one).
After restarting the IDE you should get the extra menus.
Let me know if you need any help!

hpssjellis:
> Hi Jeremy,
> I rebased the patches over latest master and added a proper example (under RPC library).
> To test this branch, the easiest way is to download the github-provided zip via https://github.com/facchinm/ArduinoCore-mbed/archive/refs/heads/portenta_flash_split.zip, and replace the files from mbed core 2.4.1 (the deprecated one).
> After restarting the IDE you should get the extra menus.
> Let me know if you need any help!

@facchinm  The install went fine, I see the extra RPC example  ```BootM4_from_SDRAM.ino``` 

Does this example just boot M4 to the SD Card or does it also re format the M7 memory to grab all the M4 memory? If it does reformat memory I don't see the relevant code. I just want to tweak the M7 and M4 memory not completely move the M4 to an SD card.

P.S. Great work so far.

Here is the example BootM4_from_SDRAM.ino


```
#include "SDRAM.h"
#include "QSPIFBlockDevice.h"
#include "MBRBlockDevice.h"
#include "FATFileSystem.h"
#include "PluggableUSBMSD.h"

QSPIFBlockDevice root(PD_11, PD_12, PF_7, PD_13,  PF_10, PG_6, QSPIF_POLARITY_MODE_1, 40000000);
mbed::MBRBlockDevice ota_data(&root, 2);
mbed::FATFileSystem ota_data_fs("fs");

void USBMSD::begin()
{
}

USBMSD MassStorage(&root);

long getFileSize(FILE *fp) {
  fseek(fp, 0, SEEK_END);
  int size = ftell(fp);
  fseek(fp, 0, SEEK_SET);

  return size;
}

void setup() {
  // put your setup code here, to run once:
  Serial.begin(115200);
  while (!Serial);

  SDRAM.begin(0);

  int err = ota_data_fs.mount(&ota_data);
  if (err) {
    Serial.println("Please run PortentaWiFiFirmwareUpdater once");
    while (1) {
      delay(10000);
    }
  }

  // Copy M4 firmware to SDRAM
  FILE* fw = fopen("/fs/fw.bin", "r");
  if (fw == NULL) {
    Serial.println("Please copy a firmware for M4 core in the PORTENTA mass storage");
    Serial.println("When done, please unmount the mass storage and reset the board");
    MassStorage.begin();
    while (1) {
      delay(10000);
    }
  }
  fread((uint8_t*)CM4_BINARY_START, getFileSize(fw), 1, fw);
  fclose(fw);

  bootM4();
}

void loop() {
  // put your main code here, to run repeatedly:
  delay(10000);
}

```

facchinm:
@hpssjellis the example shows how to perform the most "difficult" option, 100/0 split, that needs some extra steps to copy the M4 code into external SDRAM.
If you just want to tune the flash in different ways, take a look at 75/25 split that doesn't need any extra operation, just selecting the same menu entry for both M7 and M4 and then calling `bootM4()` should do the trick

hpssjellis:
> @hpssjellis the example shows how to perform the most "difficult" option, 100/0 split, that needs some extra steps to copy the M4 code into external SDRAM.
> If you just want to tune the flash in different ways, take a look at 75/25 split that doesn't need any extra operation, just selecting the same menu entry for both M7 and M4 and then calling `bootM4()` should do the trick

I didnt understand until I saw the new menu item. 




That is super easy, and truly amazing, especially the 100/0, but I am looking for finer control, instead of just:

50/50 (M7/M4)
75/25
100/0


Any chance you could add:
25/75
40/60
60/40
90/10


Or can this be done from source with whatever settings are wanted.

My use case is where one core just needs a bit more memory and the other core doesnt need as much.

hpssjellis:
@facchinm 

Just some feedback: In the regular Arduino IDE 1.8.15 when switching from M7 to M4 the "Flash Split" menu resets to ```1Mb M7 + 1Mb M4``` which is more than a little irritating.

What does seem to make sense is that I have been flashing my M7 program without bootM4(); activated, with whatever Flash Split I want, then flashing M4 with the same Flash Split, then flashing M7 with the bootM4(); function activated. So that M4 is never called before it is properly programmed.


When testing 1:1 or 75/25  (so know added complexity with booting from an SD card)
I am finding: That M4 compiles fine, M7 compiles and works fine, but M4 is having trouble booting, I get flashing red from M4, while M7 is running fine.

Is anyone else having similar issues? 

P.S. Possibly my installation is a bit messed up, it took a few trys to get it working.

P.S.S I thought M7 had a smaller flash setting anyway, since I find I can load a much larger machine learning model onto the M4 core than the M7 core. I don't really understand the setting  ```1Mb M7 + 1Mb M4 ``` since I think M4 has more flash by default than M7.  Any clarification here.

The folloswing image is some tests I did a while ago loading a much larger program onto the M4 

crashTest04

hpssjellis:
@facchinm  

Should I continue testing this Martino or wait for the next version?

hpssjellis:
@facchinm, Any updates with the split Portenta Flash? 

I am having a lot of fun with Edge Impulse loading larger Machine Learning resolution models than the regular 96x96 (ignore the skew as the OLED is 128x64), seen here  as the small rectangle showing only part of the 320x240 screen.
image


..


These models identify if the Portenta is viewing a microcontroller. The LED turns blue when it "sees" a microcontroller, and green if not.
This one is 224x224 as can be seen by the larger rectangle on screen.

image

..


When the split flash is fully working I should be able to get some much larger models loaded. Presently the split flash seems to load fine, I just can't load any larger models almost as if the larger memory is not being fully allocated.

facchinm:
Manually merged

PR #177, opened 2021-04-06
Fix incorrect signature of overloaded function

sebromero:
This PR fixes a compiler error in `Wire.h` / `Wire.cpp` where an incorrect signature was used for the overloaded function `requestFrom`.

The compiler throws the following error:
```
error: conflicting return type specified for 'virtual size_t arduino::MbedI2C::requestFrom(uint8_t, size_t)'
     virtual size_t requestFrom(uint8_t address, size_t len);
```

`size_t` was used instead of `uint8_t`

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -73,7 +73,7 @@ uint8_t arduino::MbedI2C::endTransmission(void) {
 	return endTransmission(true);
 }
 
-size_t arduino::MbedI2C::requestFrom(uint8_t address, size_t len, bool stopBit) {
+uint8_t arduino::MbedI2C::requestFrom(uint8_t address, size_t len, bool stopBit) {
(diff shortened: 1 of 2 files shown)


facchinm:
@sebromero how did you get this error?
For instance, the patch that applied this commit was https://github.com/arduino/ArduinoCore-mbed/commit/b97a5dbe7f561c5db4b52a420cdf5a116df0a5b1#diff-138e47d9e08a9bc63642821d5e4f5dd65d7912a00070cc6a5374661ecd63daa1 , and it originates from https://github.com/arduino/ArduinoCore-API/commit/4f194387a356624e8c5f4140016a69f80dc75920 .
It might be due to an outdated clone of the api repo

sebromero:
@facchinm Whoops, the API repo wasn't up to date. Closing this 😅

PR #178, opened 2021-04-06
USB host, wrapper 

marqdevx:
Copied from the keyboard controller example

This is used for a Tutorial (https://github.com/arduino/arduino-pro-content/blob/dev-lvgl2-tutorial/content/tutorials/portenta-h7/por-ard-lvgl2/content.md)

* Configure and initialize needed stuff
* Init the USB host mode
* Declare keyboard callback function
* Function to get the key that has been pressed from the keyboard


sebromero:
@marqdevx I think if we add a lightweight wrapper for mouse and keyboard I'd hide the dependency on tusb. Also I'd wrap the static functions in a class. Let's discuss this.

sebromero:
Will be replaced with new implementation

PR #209, opened 2021-04-28
Add variant for OpenMV firmware

sebromero:
This PR adds a variant for the Nano 33 BLE that allows us to expose a separate port from the OpenMV firmware. This will allow us to recognize if the board is running OpenMV without using any special tooling.
Instead of using `OpenMV` / `_omv` in the naming we could also go with `MicroPython` and `_mp`. 
Thoughts? @facchinm


facchinm:
Instead than adding another full-fledged target, I'd rather add
```
nano33ble.pid.2=0x015a
```
to the "main" entry.  This won't show the nice label but won't increase maintenance either.

facchinm (approved):

Issue #212, opened 2021-05-01
Unable to get I2C working on Raspberry Pi Pico


fsievers:
No matter what I try, I can't get I2C to work. I'm using the following i2c-scan to discover a SSD1306 display:
```
#include <Arduino.h>
#include <Wire.h>

void setup()
{
  Wire.begin();

  Serial.begin(9600);
  while (!Serial)
    ; // Leonardo: wait for serial monitor
  Serial.println("\nI2C Scanner");
}

void loop()
{
  byte error, address;
  int nDevices;

  Serial.println("Scanning...");
  
  nDevices = 0;
  for (address = 1; address < 127; address++)
  {
    Wire.beginTransmission(address);
    error = Wire.endTransmission();

    if (error == 0)
    {
      Serial.print("I2C device found at address 0x");
      if (address < 16)
        Serial.print("0");
      Serial.print(address, HEX);
      Serial.println("  !");

      nDevices++;
    }
    else if (error == 4)
    {
      Serial.print("Unknown error at address 0x");
      if (address < 16)
        Serial.print("0");
      Serial.println(address, HEX);
    }
  }
  if (nDevices == 0)
    Serial.println("No I2C devices found\n");
  else
    Serial.println("done\n");

  delay(5000); // wait 5 seconds for next scan
}
```
But I always get `No I2C devices found`. In `pins_arduino.h` it says, that SDA is on GPIO 6 and SCL is on GPIO 7 (see https://datasheets.raspberrypi.org/pico/Pico-R3-A4-Pinout.pdf). But the program is not able to discover the OLED display.

In a second try, I added
```
#define PIN_WIRE_SDA (8u)
#define PIN_WIRE_SCL (9u)
```
after the `Wire.h` include and tried to get it working on the Raspberry Pi Pico default I2C0 connections. But this doesn't work, too.

But I can confirm, that the Pico and OLED are OK and I2C is working, if I use https://github.com/earlephilhower/arduino-pico installed via Board Manager. The project uses GPIO 4 and 5 for I2C0 (Wire) by default. And I can even change the GPIOs via
```
Wire.setSDA(8);
Wire.setSCL(9);
```
in the `setup()` function right before doing `Wire.begin()`.

What can I do, to get I2C working with the ArduinoCore-mbed?

facchinm:
Hi @fsievers ,
about the I2C scan, there's a know issue with `Wire.endTransmission()` for zero bytes length packets (like the ones used during scanning).
[...]

PR #214, opened 2021-05-04
Wire: fix "Arduino way" I2C scan

facchinm:
@sievers 
Completely fixes https://github.com/arduino/ArduinoCore-mbed/issues/212

Portenta is a bit peculiar and responds correctly to the zero bytes write() but not to the read(); to be investigated.


pennam (commented):
Tested on https://github.com/arduino/ArduinoCore-mbed/commit/3a2a2486ef7a41750b1ff142622dc46f446cbe92  with 2 connected devices (LM75B and AT24CXX) read works fine also on Portenta. I think we could remove conditional compilation directives.

fsievers:
I checked the changes on my Pi Pico and can confirm, that the I2C scan is now working as expected.

PR #226, opened 2021-05-20
Correct default PDM gain for Portenta

pennam:
-1 is used to mark the gain as uninitialized then the value is set to 24.


sebromero (approved):
LGTM

PR #227, opened 2021-05-20
RP2040: avoid implicit conversion from NinaPin to int

facchinm:
Some sketches use a pattern like
```
int potpin = A5
```

Since NinaPin was an enum, the implicit cast was taking place, and any subsequent call to analogRead() would not use the Nina functions.

Works in conjunction with https://github.com/arduino-libraries/WiFiNINA/pull/180

### DIFF
--- a/cores/arduino/wiring_analog.cpp
+++ b/cores/arduino/wiring_analog.cpp
@@ -136,3 +136,8 @@ void analogReadResolution(int bits)
 {
   read_resolution = bits;
 }
+
+int getAnalogReadResolution()
+{
+  return read_resolution;
+}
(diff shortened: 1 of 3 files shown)


facchinm:
The CI correctly spotted that https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Scheduler/examples/MultipleBlinks/MultipleBlinks.ino#L28-L30 cannot work on the Nano RP2040 . What about adding a guard? Or changing the declarations into defines.

.github/workflows/compile-examples.yml
```diff
@@ -93,7 +93,8 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           fqbn: ${{ matrix.board.fqbn }}
           libraries: |
-            - name: WiFiNINA
+            - source-url: https://github.com/facchinm/WiFiNINA.git
```
aentinger, line 96: This needs to get out after https://github.com/arduino-libraries/WiFiNINA/pull/180 has been merged and the library released.

facchinm: I'll just cherry pick the first commit / force push it on this branch

PR #237, opened 2021-05-25
Improve SFU for Nano RP2040 OTA

facchinm:


aentinger (commented):
This changeset contains also the Nina Pin changes. Is this intentional?

facchinm:
> This changeset contains also the Nina Pin changes. Is this intentional?

Nope, removing :slightly_smiling_face:

libraries/SFU/src/SFU.cpp
```diff
@@ -7,7 +7,7 @@ const unsigned char SFU[0x10000] __attribute__ ((section(".second_stage_ota"), u
 	#include "rp2040.h"
 };
 
-FlashIAPBlockDevice bd(XIP_BASE + 0x100000, 0x100000);
+FlashIAPBlockDevice bd(XIP_BASE + 0xF00000, 0x100000);
```
aentinger, line 10: Do we need to have this as a static object within SFU (and therefore always on the RAM)? I'd just keep it as a local variable as done here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/257/files#diff-09aa9693675d970be15fa987b65d99e65812b4e32924fe48fce8d3f87bcd6f99R65 (Beware, I don't know if this works).

facchinm: Yes, we can indeed move it in the function or even in SFU::begin(), not sure we can see many improvements but it's worth trying

PR #258, opened 2021-06-10
Add API to WiFi class to configure wdog feed function

pennam:
This pr add the possibility to configure the watchdog feed function through API in the WiFi class and trigger the watchdog during download process.

(TODO implement timeout) -> this should go to Arduino_IoT_library i think

needs https://github.com/arduino-libraries/Arduino_Portenta_OTA/pull/10
and https://github.com/arduino-libraries/ArduinoIoTCloud/pull/261

### DIFF
--- a/libraries/WiFi/src/WiFi.h
+++ b/libraries/WiFi/src/WiFi.h
@@ -296,6 +296,9 @@ public:
 
     NetworkInterface *getNetwork();
 
+    void setFeedWatchdogFunc(voidPrtFuncPtr func);
+    void feedWatchdog();
+
 private:
 
     EMACInterface* _softAP = nullptr;
@@ -316,6 +319,7 @@ private:
     bool isVisible(const char* ssid);
     arduino::IPAddress ipAddressFromSocketAddress(SocketAddress socketAddress);
     SocketAddress socketAddressFromIpAddress(arduino::IPAddress ip, uint16_t port);
+    voidPrtFuncPtr _feed_watchdog_func = 0;
 };
 
 }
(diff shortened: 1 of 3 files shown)


libraries/WiFi/src/WiFi.h
```diff
@@ -49,6 +49,8 @@ extern "C" {
 #define DEFAULT_AP_CHANNEL 6
 #endif
 
+#define WIFI_HAS_FEED_WATCHDOG_FUNC
```
aentinger, line 52: Please use a more unique name, e.g. `ARDUINO_PORTENTA_H7_WIFI_HAS_FEED_WATCHDOG_FUNC`

libraries/WiFi/src/WiFi.h
```diff
@@ -49,6 +49,8 @@ extern "C" {
...
+
 namespace arduino {
 
 typedef void* (*voidPrtFuncPtr)(void);
```
aentinger, line 56:
```C++
typedef void(*ArduinoPortentaH7WiFiFeedWatchdogFuncPtr)(void);
```
also
```diff
-typedef void* (*voidPrtFuncPtr)(void);
+typedef void (*voidPrtFuncPtr)(void);
```

aentinger (changes requested):

libraries/WiFi/src/WiFi.h
pennam: fixed with https://github.com/arduino/ArduinoCore-mbed/pull/258/commits/28bae3e014097cf9621fc911a795c10a922e2ba7

aentinger (approved):
Tested - works :heavy_check_mark:

PR #277, opened 2021-07-05
Add support for Static Network Configuration and Socket managment

manchoz:

PR #281, opened 2021-07-09
Restructure networking libraries

facchinm:
Contains and replaces #277 and #278 
Supersedes #80 and #81 

This refactor allows a much easier porting of any network interface from mbed-os to our beloved Client/Server APIs.
Moreover, now the APIs work "as expected", so a lot of libraries that expected, for example, a non-blocking `available()`, now work again.

@hpssjellis TinyMQTT works too with this minimal patch :wink: 
```patch
diff --git a/src/TinyMqtt.h b/src/TinyMqtt.h
index b1a1cd3..6919c48 100644
--- a/src/TinyMqtt.h
+++ b/src/TinyMqtt.h
@@ -14,6 +14,10 @@
        #ifdef TCP_ASYNC
          #include <AsyncTCP.h> // https://github.com/me-no-dev/AsyncTCP
   #endif
+#elif defined(ARDUINO_PORTENTA_H7_M7)
+  #include <WiFi.h>
+  #include <WiFiClient.h>
+  #include <WiFiServer.h>
 #endif
 #ifdef EPOXY_DUINO
   #define dbg_ptr uint64_t
```


hpssjellis:
@facchinm 
Is anyone else having problems just getting the simple WiFiWebClient example working with these changes? 

Getting lots of 

```

<command-line>:0:25: warning: ISO C++11 requires whitespace after the macro name


```

Then can't find the correct WiFi library

```

Multiple libraries were found for "WiFi.h"
 Used: C:\Program Files (x86)\Arduino\libraries\WiFi
 Not used: C:\Users\Jeremy Ellis\Documents\Arduino\hardware\arduino-git\ArduinoCore-mbed-socket_async_test_pr\libraries\WiFi
 Not used: C:\Users\Jeremy Ellis\Documents\Arduino\libraries\WiFiNINA


```

PR #283, opened 2021-07-13
Restructure networking - MNX Edition

manchoz:
Add a few missing methods and fixes mutex handling.
Refactor AdvancedChatServer example.


manchoz:
Refactor copy constructor and assignment operator.
Restore classic AdvacedChatServer.

PR #294, opened 2021-07-25
Enable 320x320 resolution for himax HM01B0

idohalamit:
I did that the same way OpenMV did:
github.com/openmv/openmv/blob/master/src/omv/sensors/hm01b0.c

 - Added constants consistently with QVGA and QQVGA modes.

 - Updated himax.cpp, camera.cpp accordingly.

My first commit to an open source project <3

### DIFF
--- a/libraries/Portenta_Camera/camera.cpp
+++ b/libraries/Portenta_Camera/camera.cpp
@@ -461,7 +461,7 @@ void CameraClass::HIMAXIrqHandler()
 
 int CameraClass::begin(uint32_t resolution, uint32_t framerate)
 {  
-  if (resolution > CAMERA_RMAX) {
+  if (resolution >= CAMERA_RMAX) {
     return -1;
   }
 
--- a/libraries/Portenta_Camera/camera.cpp
+++ b/libraries/Portenta_Camera/camera.cpp
@@ -10,6 +10,7 @@
 static const int CamRes[][2] = {
     {160, 120},
     {320, 240},
+    {320, 320},
 };
 static __IO uint32_t camera_frame_ready = 0;
 static md_callback_t user_md_callback = NULL;
@@ -460,7 +461,7 @@ void CameraClass::HIMAXIrqHandler()
 
 int CameraClass::begin(uint32_t resolution, uint32_t framerate)
 {  
-  if (resolution >= CAMERA_RMAX) {
+  if (resolution > CAMERA_RMAX) {
     return -1;
   }
 
--- a/libraries/Portenta_Camera/camera.h
+++ b/libraries/Portenta_Camera/camera.h
@@ -3,7 +3,8 @@
 enum {
     CAMERA_R160x120 = 0x00,   /* QQVGA Resolution   */
     CAMERA_R320x240 = 0x01,   /* QVGA Resolution    */
-    CAMERA_RMAX     = 0x02    /* 320x320 Resolution */
+    CAMERA_R320x320 = 0x02,    /* 320x320 Resolution */
+    CAMERA_RMAX
 };
 
 typedef void (*md_callback_t)();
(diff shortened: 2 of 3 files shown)


facchinm:
Hi @idohalamit , thanks for your submission! I'm reviewing the code and there are a bunch of small things that need fixing but overall it's good! I'm commenting the needed changes inline, let me know if they sound ok to you  :wink:

libraries/Portenta_Camera/camera.cpp
```diff
@@ -460,7 +461,7 @@ void CameraClass::HIMAXIrqHandler()
 
 int CameraClass::begin(uint32_t resolution, uint32_t framerate)
 {  
-  if (resolution >= CAMERA_RMAX) {
+  if (resolution > CAMERA_RMAX) {
```
facchinm, line 464: No need to do this if we add a `CAMERA_R320x320` entry to the enum

libraries/Portenta_Camera/camera.h
```diff
@@ -1,9 +1,9 @@
 #include "drivers/InterruptIn.h"
 
 enum {
-    CAMERA_R160x120 = 0x00,   /* QQVGA Resolution */
-    CAMERA_R320x240 = 0x01,   /* QVGA Resolution  */
-    CAMERA_RMAX
+    CAMERA_R160x120 = 0x00,   /* QQVGA Resolution   */
+    CAMERA_R320x240 = 0x01,   /* QVGA Resolution    */
+    CAMERA_RMAX     = 0x02    /* 320x320 Resolution */
```
facchinm, line 6: Let's add a `CAMERA_R320x320` mode and keep `CAMERA_RMAX` without any attached value, since it's used in a  "last element of the enum" fashion.

facchinm (changes requested):

libraries/Portenta_Camera/camera.h
idohalamit: Looks good!

facchinm: Not really what I meant :slightly_smiling_face: 
What I'm asking for is something like 
```
enum {
    CAMERA_R160x120 = 0x00,   /* QQVGA Resolution */
    CAMERA_R320x240 = 0x01,   /* QVGA Resolution  */
    CAMERA_R320x320 = 0x02,   /* 320x320 Resolution */
    CAMERA_RMAX
}
```
then, use the `CAMERA_R320x320` as reference.

sebromero:
```suggestion
    CAMERA_R320x320     = 0x02,    /* 320x320 Resolution */
    CAMERA_RMAX
```

idohalamit:
Should I commit anything more?

sebromero:
@idohalamit I think we're good. @facchinm If you agree I suggest we merge 🚢

sebromero (approved):
LGTM

idohalamit:
I do!
I had some trouble with the serial connection in the [demo,](https://docs.arduino.cc/tutorials/portenta-vision-shield/vs-ard-gs) I did some workaround.
They should work even though I can't test them right now.
Also, you'd probably want to change the hard-coded serial port.
[LiveFeed320x320.ino.txt](https://github.com/arduino/ArduinoCore-mbed/files/6988619/LiveFeed320x320.ino.txt)
[CameraViewer320x320.pde.txt](https://github.com/arduino/ArduinoCore-mbed/files/6988622/CameraViewer320x320.pde.txt)

facchinm:
Merged, thanks a lot for your contribution :slightly_smiling_face: !

sebromero:
@idohalamit Thanks. Yes, the example sketch for this has some issues on Windows. I haven't gotten around to investigate as I don't have a Windows computer. Do you have an idea what exactly gets messed up in regards to the serial connection?

idohalamit:
I haven't isolated the problem, but sending the data in small chunks (I used 512 bytes), using Serial.flush() and adding delays between sends fixes it.

PR #331, opened 2021-09-22
Fix ns_cmdline spilling `cmd` variable

### LINKED ISSUE #310: Cannot create a variable called "cmd"
I found something weird. My sketch, which has a variable called "cmd", will not compile on the pi pico.
It throws this compile error:
.../Arduino15/packages/arduino/hardware/mbed_rp2040/2.4.1/variants/RASPBERRY_PI_PICO/libs/libmbed.a(ns_cmdline.o):(.bss.cmd+0x0): multiple definition of `cmd'

Even this simple sketch fails:
```c++
char cmd;

void setup() {
}

void loop() {
}
```

facchinm:
FIxes #310


facchinm:
Manually merged

PR #338, opened 2021-09-29
Portenta H7 3x I2C bus

RoboCode23:
The Portenta has 3x I2C bus and only 2 are defined in the libraries...

added and modified howmany from 2 to 3 in pins_arduino.h
``` 
#define WIRE_HOWMANY		3
#define I2C_SDA2			(PH_12)
#define I2C_SCL2			(PH_11)
```
and added into Wire.h and Wire.cpp
```
#if WIRE_HOWMANY > 2
arduino::MbedI2C Wire2(I2C_SDA2, I2C_SCL2);
#endif
```

makes all 3 work


facchinm:
Not the right way to submit a PR unfortunately :slightly_smiling_face: 
If you want you can submit it via Github interface by navigating to the variant file, then click the pencil on the top right, apply your modification and that's it.

Issue #194, opened 2021-04-14
Raspberry Pico SPI pins


sstaub:
Raspberry Pico
How to determine which pins are used for SPI, and how to do this for SPI1. Are there any standard pins?
Same for I2C. Is there any API (with examples)?

Grumpy-Mike:
[Pico-R3-A4-Pinout.pdf](https://github.com/arduino/ArduinoCore-mbed/files/6315485/Pico-R3-A4-Pinout.pdf)
You can use any of these pins for I2C, note there are two I2C buses as well. You have to tell the Pico what pins you are going to use when you initialise it.

sstaub:
I know about the pins I can use. But how to configure?

Grumpy-Mike:
This is an example of using the I2C to drive an LCD

[lcd_1602_i2c.c.zip](https://github.com/arduino/ArduinoCore-mbed/files/6315685/lcd_1602_i2c.c.zip)

sstaub:
@Grumpy-Mike What have the code todo with ArduinoCore-Core?

Grumpy-Mike:
Don't understand the question sorry.

stuartdd:
This is my understanding | miss-understanding.

I have been using the code from earlephilhower/arduino-pico with some success.
       My SSD1306 plays Space Invaders, nice!

I wanted to use the official Arduino release and followed the link to this repo.

Initial Process (Not using README.md):
1) I downloaded a fresh Arduino IDE, Removed all Arduino related libs from my user and started the IDE.
2) I went to the Board Manager an installed 'Arduino OS RP2040 Boards'.
3) I selected the 'Raspberry PI Pico' board.
4) I uploaded the Blink sketch pressed upload (which failed)
5) I Ran the post-install.sh (from your repo) as root. This fixed the above issue!
6) I uploaded the sketch and the LED started to blink. YAHAOO!
        Does this use the code from ArduinoCore-mbed?
        If NOT what is the difference between ArduinoCore-mbed and the IDE's Board installation?
7) I loaded up a sketch that uses the I2C interface to drive a OLED display using 'Adafruit_SSD1306.h' library.
8) I cannot find a way of defining the SDA and SCL pins used by the Wire api. 
9) I looked at lcd_1602_i2c.c.zip and it says:
   GPIO 4 (pin 6)-> SDA on LCD bridge board
   GPIO 5 (pin 7)-> SCL on LCD bridge board
[...]

PR #393, opened 2022-01-24
Change the default SPI & I2C pins to match pinout docs

TedPap:
This PR changes the default pins for SPI and I2C to match the ones marked as default in the [official raspberry pi pico pinout](https://datasheets.raspberrypi.com/pico/Pico-R3-A4-Pinout.pdf).

This has already been discussed in issue #194 but I couldn't understand if this change was desired or not but I decided to submit it anyways.

I've tested it with I2C and it works. Unfortunately I could not test with SPI.

Please excuse me if I'm missing something when it comes to PRs in this repository. Finally, thank you very much for your hard work on this project.

### DIFF
--- a/variants/RASPBERRY_PI_PICO/pins_arduino.h
+++ b/variants/RASPBERRY_PI_PICO/pins_arduino.h
@@ -42,10 +42,10 @@ static const uint8_t A3  = PIN_A3;
 #define PIN_SERIAL_RX (1ul)
 
 // SPI
-#define PIN_SPI_MISO  (4u)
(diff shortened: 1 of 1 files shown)


facchinm:
It makes a lot of sense since the documentation now explicitly states they are the "default" port. Thanks!

TedPap:
Awesome! Thanks for merging this @facchinm. Any idea when the next release is going to be?

salmanfarisvp:
For others: The default I2C is on 
GPIO4 -> SDA
GPIO 5 -> SCL

PR #396, opened 2022-01-27
Fix incorrect OMV PID

sebromero:
The PID listed in the boards file does not correspond that what is exposed through the OMV firmware. This PR adjusts that. Another option would be to adjust it on the OMV side. @facchinm Thoughts?

Screenshot 2022-01-27 at 10 31 19


sebromero:
@facchinm I repurposed the PR to add some documentation to the PIDs :-)

PR #399, opened 2022-02-07
Add WiFI.config() functionality to mbed Wifi Library

SrJoosh:
I noticed that WiFi.config does not exist in the wifi library on this board to setup a static ip.

Added:
WiFi.Config(IPV4,SUBNET,GATEWAY); function back to set the network address and disable DHCP for static station mode wifi setup.


aentinger (approved):
LGTM :+1:

PR #401, opened 2022-02-10
Add GPS to exported libraries

manchoz:
Export the GPS library too


manchoz:
@facchinm @Rocketct please, review.

PR #402, opened 2022-02-15
`Ethernet::MACAddress()` implementation

d-a-v:
Use Mbed's `macAddress()` implementation for Arduino `MACAddress()`

The Mac Address seems to be stable/valid only after `::begin()`,
However, I can't explain myself why my Portenta H7's OUI `2B:00:1B` cannot be found on the various OUI databases available on internet.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -84,4 +84,9 @@ unsigned long arduino::EthernetClass::getTime() {
   return 0;
 }
 
+void arduino::EthernetClass::MACAddress(uint8_t *mac_address)
+{
+  macAddress(mac_address);
+}
+
 arduino::EthernetClass Ethernet;


d-a-v:
> you can already obtain the mac address with `Ethernet.macAddress(uint8_t*)`

The purpose of this PR was to use this very method to implement missing code for [`Ethernet.MACAddress()`](https://www.arduino.cc/en/Reference/EthernetMACAddress). On this Arduino API documentation, the official example can be found but it cannot be linked @facchinm.

This pull-request permits to avoid the `undefined reference` error when one tries to use the official Arduino API.

I assume you missed the fact that this is a pull-request and not an issue, or I missed something else in your closing answer, and would be pleased to be enlightened :)

facchinm:
Oh well, I thought I knew everything about Ethernet class but I was wrong :slightly_smiling_face: Thanks for pointing it out.
The old `MACAddress()` method was indeed replaced with `macAddress()` on all libraries following the original Ethernet one, but since this one is still there and documented I'll keep it (even without adding the `deprecated` tag)

PR #404, opened 2022-02-17
Change digitalPinToPinName() into a first class function

facchinm:
This fixes some issues with the new arduino-cli, since compiling a lot of libraries forces linking them as archives, breaking the inline.

Should fix https://github.com/arduino-libraries/Arduino_OV767X/pull/20
@rajames


rajames:
This solves the issue. Thanks

PR #405, opened 2022-02-18
[WIP] Manage TLS Server Name Indication

manchoz:
Enable Server Name Indication by default for TLS connections (recommended and mandatory configuration) and add a method for explicitly disabling it.


manchoz:
Tested working in different scenarios.

@facchinm, please, review.

Issue #424, opened 2022-03-13
Example sketch `USBHID>Keyboard` makes the Arduino unusable


beyarkay:
## How the problem started
I was playing around with using the Arduino Nano BLE 33 as a keyboard, and stumbled upon the [USBHID>Keyboard sketch](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/USBHID/examples/Keyboard/Keyboard.ino) which I got via the default suggestion from the Board Manager to install `Arduino Mbed OS Nano Boards`:

See the source reproduced below:

```c
#include "PluggableUSBHID.h"
#include "USBKeyboard.h"

USBKeyboard Keyboard;

void setup() {
  // put your setup code here, to run once:

}

void loop() {
  // put your main code here, to run repeatedly:
  delay(1000);
  Keyboard.printf("Hello world\n\r");
}
```

It seemed fairly innocent, so I uploaded it without a second thought thinking that I could trust anything provided from Arduino itself as an example script.

## What the problem is
But now, as soon as I plug in the Nano, it types `Hello world` followed by a newline every. single. second.

If I try to upload the blink sketch to clear it, I first need to select the Port, but the act of going through the menu to choose the correct port takes over a second, and typing in `Hello world` selects the library manager and ends up opening it when enter is pressed due to the `\n\r`. So my Arduino Nano has been rendered useless by code included as an example in the default sketches. That's not cool.

## Failed Solution Number 1
I tried creating a custom keyboard shortcut to select the port in less than a second, and then pressing `cmd-U` to upload, but I think there might also be a problem there because I get this error:

```
    ...
    Found upload port: /dev/cu.usbmodem11401
    /Users/<my username>/Library/Arduino15/packages/arduino/tools/bossac/1.9.1-arduino2/bossac -d --port=cu.usbmodem11401 -U -i -e -w /var/folders/y1/md9_k_td4rz1f0pq26d24crh0000gn/T/arduino_build_230227/<my script name>.ino.bin -R 
->  No device found on cu.usbmodem11401
    An error occurred while uploading the sketch
    ....
```

I'm absolutely certain that `cu.usbmodem11401` is the correct port, because that's what shows up when I `ls /dev/cu*` while the Arduino is plugged in. So I think that the Arduino IDE is attempting to upload to `cu.usbmodem11401`, but sees that it's a keyboard, and decides that it can't upload code to a keyboard so aborts.

## Failed Solution Number 2
[...]

PR #425, opened 2022-03-13
Fix: short delay prevented new code uploads.

beyarkay:
When the Arduino is treated as a keyboard and automatically typed `Hello world\n\r` every second, it was nearly impossible to upload new code due to the typing interfering with manually choosing the correct port. See [this issue](https://github.com/arduino/ArduinoCore-mbed/issues/424) for more details.

Mitigates [this issue](https://github.com/arduino/ArduinoCore-mbed/issues/424).


beyarkay:
Closing as it seems this will not be merged. Because who wants to fix things when they can just stay broken

PR #431, opened 2022-03-23
Add support for secure bootloader

pennam:


pennam:
CI failing due to missing imgtool binaries

PR #439, opened 2022-04-01
Drastically speedup client.read() operation

### LINKED ISSUE #438: [Portenta_H7] Ethernet WebServer extremely slow from core v2.7.2 - v3.0.1
#### Describe the bug

The core's Ethernet Server speed is normal for core v2.6.1-

For `core v2.7.2 to v3.0.1`, the Ethernet Server speed is **extremely slow and almost unusable**.

For example, try from complex example [**AdvancedWebServer**](https://github.com/khoih-prog/EthernetWebServer/tree/master/examples/AdvancedWebServer) to the simplest one [**WebServer**](https://github.com/arduino/ArduinoCore-mbed/tree/master/libraries/Ethernet/examples/WebServer)

#### MRE

Using the following core's Ethernet library example [**WebServer**](https://github.com/arduino/ArduinoCore-mbed/tree/master/libraries/Ethernet/examples/WebServer)  with 

- Arduino IDE v1.8.19
- mbed_portenta `core v2.6.1` or `core v2.7.2 to v3.0.1`
- Portenta_H7_M7
- Either static IP or DHCP

#### Related issue
[**[Portenta_H7] Ethernet not working in core v3.0.0** #436](https://github.com/arduino/ArduinoCore-mbed/issues/436)

facchinm:
Since the data being consumed by the reader thread is going to be read somewhere via read() APIs (that are known to be fair and call yield()), we can avoid the delay() when the buffer needs to be flushed by setting the reader thread priority to something lower than the main thread.

To be tested on some real world applications (by now, tested with Ethernet/WiFi Client/Server "basic" examples)

Fixes #438 

@manchoz

### DIFF
--- a/libraries/SocketWrapper/src/MbedClient.cpp
+++ b/libraries/SocketWrapper/src/MbedClient.cpp
@@ -213,7 +213,7 @@ size_t arduino::MbedClient::write(const uint8_t *buf, size_t size) {
   int ret = NSAPI_ERROR_WOULD_BLOCK;
   do {
     ret = sock->send(buf, size);
-  } while (ret != size && connected());
+  } while ((ret != size && ret == NSAPI_ERROR_WOULD_BLOCK) && connected());
   configureSocket(sock);
   return size;
 }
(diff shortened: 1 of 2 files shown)


giulcioffi (approved):
Tested with Ethernet:
- without the patch: 187040 bytes in 75.6310, rate = 2.47 kbytes/second
- with the patch: 181118 bytes in 3.9838, rate = 45.46 kbytes/second

Issue #437, opened 2022-03-29
[RP2040] Built in mbedtls does not have ctr_drbg and entropy enabled


dogtopus:
I'm trying to use the built-in mbedtls (on RP2040) to do RSA PSS signing, which requires an RNG function. Unfortunately I found that when I try to use  ctr_drbg and entropy API the linker gives me the "undefined reference" error.

It also seems hard to workaround this issue by using external mbedtls due to symbol conflicts. I'd also avoid switching to another crypto library at this point due to the fact that I already wrote a bunch of mbedtls specific code.

dogtopus:
It seems to be worse than I initially thought. IIRC some versions of TLS uses RSA PSS so I ran nm on all the the libmbed.a targeting supported boards, and found that only RP2040 have those APIs missing, and looks like they also don't have SSL/TLS support. This includes Nano RP2040 connect which has onboard wifi.

facchinm:
Hi @dogtopus ,
sorry for the inconvenience. Can you paste a snippet that triggers the  "undefined reference" error so I can tune our mbedtls config for RP2040 and get it included? Thanks

facchinm:
Nano RP2040 uses the onboard ESP32 module to perform TLS so it doesn't use any mbedtls functionality "directly"

dogtopus:
What I have is currently somewhat like this:

```c++
#include <mbedtls/entropy.h>
#include <mbedtls/ctr_drbg.h>

static mbedtls_ctr_drbg_context drbg;
static mbedtls_entropy_context entropy;
//static mbedtls_rsa_context rsa;

const uint8_t DRBG_CUSTOM[] = "CUSTOM";

int my_entropy_source(void *ctx, uint8_t *buf, size_t buflen, size_t *outlen) {
    // TODO implement this
    return MBEDTLS_ERR_ENTROPY_SOURCE_FAILED;
}

bool initRNG() {
    int mbedtlsError;

    mbedtls_entropy_init(&entropy);
    mbedtlsError = mbedtls_entropy_add_source(&entropy, my_entropy_source, nullptr, 32, MBEDTLS_ENTROPY_SOURCE_STRONG);
    if (mbedtlsError != 0) {
        Serial.print("mbedtls_entropy_add_source() -> ");
        Serial.println(mbedtlsError, 16);
        return false;
    }

    mbedtls_ctr_drbg_init(&drbg);
[...]

PR #445, opened 2022-04-04
Add entropy source for Nano RP2040 Connect

facchinm:
Fixes https://github.com/arduino/ArduinoCore-mbed/issues/437 in RP2040 based boards

@dogtopus may you give it a spin?


dogtopus:
I don't have Nano RP2040 Connect so I ported the change to RPi Pico and gave it a test. The signing function now happily accepts the random seed instead of just returning an error code (as seen when I use a stub RNG function that fills the buffer with 0). So I guess I'll call this a pass?

Also could you please also add this change to Pico?

PR #447, opened 2022-04-07
Rename MCUboot default keys to make more clear how they are used

pennam:
`ecdsa-p256-signing-priv-key.pem` is used by imgtool to sign the image while the [public dump](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/STM32H747_System/examples/STM32H747_updateBootloader/ecdsa-p256-signing-key.h) included in the bootloder binary is used for image verification.

`ecdsa-p256-encrypt-pub-key.pem` is used by imgtool for image encryption  with elliptic curve integrated encryption scheme and the [private dump](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/STM32H747_System/examples/STM32H747_updateBootloader/ecdsa-p256-encrypt-key.h) included in the bootloader binary is used to decrypt the image.


umbynos:
Maybe could be good to also update [this](https://arduino.github.io/arduino-cli/dev/guides/secure-boot/)? I can do that if you are not comfortable with that.

PR #456, opened 2022-04-14
Fix RAM byte in Nicla Vision BL descriptor

giulcioffi:
In Nicla Vision BL descriptor the byte corresponding to the external RAM size was incorrectly set to 8. External RAM is missing in Nicla Vision board.
The binaries have been fixed and the bootloader sketches in STM32H747_System library have been updated to print `RAM size: N/A` if the RAM size byte is equal to 0.


sebromero (approved):
LGTM

PR #458, opened 2022-04-16
Fix unsigned int overload: write(0u)

nimish:
`Wire::write(0u)` is otherwise ambiguous in modern C++ compilers, introduced by commit 2958bc00a985cbe5b63400fb4afa117453750255

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -57,10 +57,10 @@ class MbedI2C : public HardwareI2C
     virtual void onRequest(void(*)(void));
 
     virtual size_t write(uint8_t data);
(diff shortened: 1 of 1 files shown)


libraries/Wire/Wire.h
```diff
@@ -57,7 +57,10 @@ class MbedI2C : public HardwareI2C
     virtual void onRequest(void(*)(void));
 
     virtual size_t write(uint8_t data);
-    virtual size_t write(int data) {
+    virtual size_t write(intmax_t data) {
+      return write ((uint8_t)data);
```
rmlearney, line 61: Should this be int8_t?

PR #472, opened 2022-05-01
PDM won't terminate executing program

markub3327:
Hi,

   this is the non-terminating version of original PDM library for Arduino Nano RP2040 Connect. It's not good if system crash when a microphone buffer is full. The better way is signalising it and continue executing program. Data that weren't read trashed.

I test it with my SD card Audio Recorder and works better like original code. I can place an example of it here too.

Thanks.

### DIFF
--- a/libraries/PDM/src/rp2040/PDM.cpp
+++ b/libraries/PDM/src/rp2040/PDM.cpp
@@ -14,7 +14,7 @@ extern "C" {
 #include "pdm.pio.h"
 
 // Hardware peripherals used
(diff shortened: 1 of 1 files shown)


markub3327:
The better way of defining `dmaChannel` is:
```python
const uint dmaChannel = dma_claim_unused_channel(true);
```

If you're using the more DMA channels the default channel 0 may be used by another device !

facchinm:
Looks like a sensible change, thanks for posting! merging right now

PR #475, opened 2022-05-06
Rename STM32H747_updateBootloader.ino -> STM32H747_manageBootloader.ino

pennam:
The sketch can be used not only to update but also to switch between the two bootloaders available.


pennam:
@umbynos @ubidefeo if we go for this i think we need to update also the secure boot tutorial

ubidefeo (approved):
LGTM

PR #481, opened 2022-05-21
Fix pico.upload.maximum_size

tttapa:
Even though the RP2040 supports up to 16 MiB of flash, the Pi Pico only has 2 MiB installed. (<https://www.raspberrypi.com/products/raspberry-pi-pico/specifications>)

This change also fixes the percentage of program storage space in use reported by the builder/IDE.

The linker file does specify the correct size: 
https://github.com/arduino/ArduinoCore-mbed/blob/8d5510ad677ff84dcfbc0fe9626c9d1cb4329763/variants/RASPBERRY_PI_PICO/linker_script.ld#L3

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -414,7 +414,7 @@ pico.upload.transport=
 pico.upload.use_1200bps_touch=true
 pico.upload.wait_for_upload_port=false
 pico.upload.native_usb=true
-pico.upload.maximum_size=16777216
+pico.upload.maximum_size=2097152
 pico.upload.maximum_data_size=270336
 
 pico.bootloader.tool=openocd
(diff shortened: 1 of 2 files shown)


facchinm:
Hi @tttapa ,
thanks for spotting this!
Can you sign the CLA so I can merge the patch directly? Thanks

tttapa:
Related to this issue, it seems that some sections (e.g. `.rodata`) that end up in flash are not counted in the total. If you use a lot of constants and string literals, the reported memory usage is significantly lower than the actual usage.  
It might be worth looking into changing the regex here:
https://github.com/arduino/ArduinoCore-mbed/blob/2d4c3073d224a504ce68920bcccef71109d1cabc/platform.txt#L104
```txt
recipe.size.regex=^(?:\.data|\.text|\.rodata)\S*?\s+([0-9]+).*
```

>  Can you sign the CLA so I can merge the patch directly? Thanks

I doubt I can claim any rights to the number 2097152 :)

If that's okay, I would prefer not to share my private GitHub profile data with cla-assistent.io, but I do agree to the CLA.

facchinm:
Both patches applied, thank you!

Issue #504, opened 2022-06-20
Feature Request: PortentaH7 Camera working using M4 


hpssjellis:
@facchinm (I wouldn't bug you if it wasn't important!)

Someone on the [edgeImpulse.com forum](https://forum.edgeimpulse.com/t/object-detection-on-m4-core/4778) is trying to get a machine learning vision model working on the M4 core. When I try even the simple Arduino [camera raw bytes](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Camera/examples/CameraCaptureRawBytes/CameraCaptureRawBytes.ino) example I get the same results as a complex sketch:

```
C:\Arduino-1.8.19\portable\packages\arduino\hardware\mbed_portenta\3.1.1\libraries\Camera\src\camera.cpp: In function 'void HAL_TIM_PWM_MspInit(TIM_HandleTypeDef*)':
C:\Arduino-1.8.19\portable\packages\arduino\hardware\mbed_portenta\3.1.1\libraries\Camera\src\camera.cpp:120:27: error: 'DCMI_TIM' was not declared in this scope
     if (htim->Instance == DCMI_TIM) {


```

Is this something that might have an easy fix or is it more complex than that?

facchinm:
Hi Jeremy,
this is due to [this](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Camera/src/camera.cpp#L31) and [this](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Camera/src/camera.cpp#L68) lines, that should read `#if defined(PORTENTA_H7_M7) || defined(PORTENTA_H7_M4)` to apply to M4 core too.
Let me know if changing it makes the camera work on the M4, because there may be other issues with memory zones and DMA. If it works, I'll be glad to merge the patch immadiately

hpssjellis:
Thanks @facchinm . I spent the whole day on it, summarized in the README.md file [here](https://github.com/hpssjellis/my-examples-for-the-arduino-portentaH7/tree/master/vision-shield/camera-m4)

Basically ```Camera::grabFrame(``` needs ```SCB_InvalidateDCache_by_Addr(```, but I can't seem to get it included for the PortentaH7 M4 core, it runs fine on the M7 Core. 

[...]

PR #512, opened 2022-06-23
PR: update camera.cpp for the PortentaM4 Core

hpssjellis:
As per this issue [Feature Request: PortentaH7 Camera working using M4 #504](https://github.com/arduino/ArduinoCore-mbed/issues/504) and great suggestions from @facchinm I managed to get the Portenta M4 Camera working by making a few changes to the camera.cpp file. Specifcally adding 4 sets of 

```

#if defined (ARDUINO_PORTENTA_H7_M7) || defined (ARDUINO_PORTENTA_H7_M4)

```

The only hack I am a bit concerned about is [here](https://github.com/hpssjellis/ArduinoCore-mbed/blob/1c3150fa382689bdce1f4d483dff033328dea0a8/libraries/Camera/src/camera.cpp#L608-L611) where I excluded the following function using an ```#ifndef```

```

    #ifndef ARDUINO_PORTENTA_H7_M4  // do not invalidate if Portenta M4 Core
       // Invalidate buffer after DMA transfer.
       SCB_InvalidateDCache_by_Addr((uint32_t*) framebuffer, framesize);
    #endif

```

Not really sure if excluding the ```SCB_InvalidateDCache_by_Addr```  is the best solution for the M4 core.

I have tested the PR on both an [OLED program](https://github.com/hpssjellis/portenta-pro-community-solutions/blob/main/examples/dot3-portenta-vision-shields/dot35-camera-and-oled/dot351-camera-oled-regular/dot351-camera-oled-regular.ino) and an [edgeImpulse 64x64 FOMO Vision](https://github.com/hpssjellis/portenta-pro-community-solutions/tree/main/examples/dot5-portenta-machine-learning/dot51-portenta-edge-impulse-ml/dot514-vision-fomo/dot5146-oled-count) program as well as the standard [CameraMotionDetect](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Camera/examples/CameraMotionDetect/CameraMotionDetect.ino)  (flashes blue LED on motion.)

### DIFF
--- a/libraries/Camera/src/camera.cpp
+++ b/libraries/Camera/src/camera.cpp
@@ -605,7 +605,7 @@ int Camera::grabFrame(FrameBuffer &fb, uint32_t timeout)
 
     HAL_DCMI_Suspend(&hdcmi);
 
-    #ifndef ARDUINO_PORTENTA_H7_M4  // do not invalidate if Portenta M4 Core
+    #if defined(__CORTEX_M7)  // only invalidate buffer for Cortex M7
        // Invalidate buffer after DMA transfer.
        SCB_InvalidateDCache_by_Addr((uint32_t*) framebuffer, framesize);
     #endif
--- a/libraries/Camera/src/camera.cpp
+++ b/libraries/Camera/src/camera.cpp
@@ -28,7 +28,7 @@
 #define ALIGN_PTR(p,a)   ((p & (a-1)) ?(((uintptr_t)p + a) & ~(uintptr_t)(a-1)) : p)
 
 // Include all image sensor drivers here.
-#ifdef ARDUINO_PORTENTA_H7_M7
+#if defined (ARDUINO_PORTENTA_H7_M7) || defined (ARDUINO_PORTENTA_H7_M4)
 
 #define DCMI_TIM                    (TIM1)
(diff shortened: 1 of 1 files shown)


libraries/Camera/src/camera.cpp
```diff
@@ -605,9 +605,10 @@ int Camera::grabFrame(FrameBuffer &fb, uint32_t timeout)
 
     HAL_DCMI_Suspend(&hdcmi);
 
-    // Invalidate buffer after DMA transfer.
-    SCB_InvalidateDCache_by_Addr((uint32_t*) framebuffer, framesize);
-
+    #ifndef ARDUINO_PORTENTA_H7_M4  // do not invalidate if Portenta M4 Core
```
facchinm, line 608: I'd personally replace this with ```#if defined(__CORTEX_M7)``` since data/instruction caches only exists for M7 and it will be portable to Nicla Vision eventually without no modifications.

facchinm (changes requested):

libraries/Camera/src/camera.cpp
facchinm: I think it's fine, Nicla Vision (not Sense) needs a bit of work to get the example up and running but I can assure that the patch is correct :slightly_smiling_face:

PR #531, opened 2022-08-01
MbedClient: retrieve the address and port from the underlying socket …

romainreignier:
…and implement remotePort()

Useful for Client retrieved from Server::available() and tested with client example


facchinm:
LGTM! Thanks you very much, merging

PR #524, opened 2022-07-20
Call objcopy.postobjcopy only when needed

facchinm:

PR #533, opened 2022-08-09
Remove imgtool dependency from not supported platforms

pennam:
After #524 has been merged we can remove `imgtool` dependency from platforms not using it

/cc @umbynos


umbynos (approved):

pennam:
> I suggest also to bump to the [latest version of imgtool](https://github.com/arduino/imgtool-packing/releases/tag/1.8.0-arduino.2) which fixes the mac binary not placed in a top level directory

:+1: See #534

PR #535, opened 2022-08-13
Always clean up previous I2C object before new to avoid leaking memory

waynepiekarski:
I found a memory leak while debugging some code that was calling Wire.begin() repeatedly. The Wire implementation currently calls new each time but does not free up any previously allocated memory. Most other Wire.h implementations I've seen don't allocate any dynamic memory so calling begin() multiple times is not a problem.

This patch makes sure to call end() first to clean up any previous I2C object if one exists.


facchinm:
Thanks you for the contribution @waynepiekarski ! Calling repeatedly `Wire.begin()` without `end()` is clearly not a best practice but the overhead fro this patch is minimal so I think we can safely merge it.

PR #548, opened 2022-09-07
User friendliness

Epipeppy:
Step 1) Adding control system (pin and if statement).

### DIFF
--- a/libraries/USBHID/examples/Keyboard/Keyboard.ino
+++ b/libraries/USBHID/examples/Keyboard/Keyboard.ino
@@ -1,15 +1,26 @@
+/**************************************************************
+** Brief introduction to a few HID Keyboard commands.        **
+** Unlike the other Keyboard library, this one doesn't use   **
+** the "press", "release", and "send" commands.              **
+**************************************************************/
+
 #include "PluggableUSBHID.h"
 #include "USBKeyboard.h"
 
 USBKeyboard Keyboard;
 
-void setup() {
-  // put your setup code here, to run once:
+// Arbitrary pin.
+const int CONTROL_PIN = 2;
 
+void setup() {
+  // We will use this to start/end the prints.
+  pinMode(CONTROL_PIN, INPUT_PULLUP);
 }
 
 void loop() {
-  // put your main code here, to run repeatedly:
-  delay(1000);
-  Keyboard.printf("Hello world\n\r");
+  // This will run only if the control pin is connected to ground.
+  if( digitalRead(CONTROL_PIN) == LOW){
+    delay(1000);
+    Keyboard.printf("Hello world\n\r");
+  }
 }


sebromero:
Hi @Epipeppy ! Thank you for your contribution. As far as I understand you would like to avoid that the string is printed over and over. What if we just move the code to the setup function instead? That way we can avoid that the people need to wire up any external component to test this.

Epipeppy:
> Hi @Epipeppy ! Thank you for your contribution. As far as I understand you would like to avoid that the string is printed over and over. What if we just move the code to the setup function instead? That way we can avoid that the people need to wire up any external component to test this.

Hello, and sorry in advance for the upcoming wall of text.

My main issue was that after uploading the sketch, I was no longer able to connect to and reprogram my board without doing a hard reset (? double tap reset key); which seeemed extreme for a simple example program.  Secondary was the fact that the program was sending the keys to my computer repeatedly, which was messing with any program I had open that was focused (i.e. not in the background).  A tertiary concern was the fact that the example didn't show how to use the Keyboard library for mbed-OS; for example, the `printf()` function isn't defined in the Keyboard library but was used.  The normal Keyboard library also used functions like `press()`, `release()`, and `send()`; however the functions that are defined in the mbed-OS are (or were at the time) `key_code()​​​`, and `_putc()​​​`.

Initially I wrote a fully reworked example to show how to use these, but was told I should post them in small chunks (small changes), rahter than 1 big change, so that any issues could be more easily resolved.  From some of the other posts or comments I read, this example, when done with other boards, usually has a "button" or something to enable the output; that way the output is only sent when the button is pressed.  It seemed like a useful feature, because of my secondary issue, so I incorporated it.  If you have an idea for something we could put in the setup loop that would solve this, I say that sounds great! 

I'm happy to work more on this if that would be helpful.  I remember feeling a bit frustrated at the time, and I wanted (and still want) to help others avoid that.

Epipeppy:
Re-reading my reply, I noticed that I probably misunderstood what you said.  You said put the print in the setup function (which I called the setup loop for some reason), that way it would only print once and be easily testable.  That's a pretty good idea which would solve one of the issues I was listing.  

In regards to showing how to use the library, do you think creating a different example sketch would be a better solution than heavily modifying this one?

Epipeppy:
Hello @alessandromrc debouncing isn't a bad idea; I personally just connect the GPIO to ground via wire, and leave it unplugged when I'm not testing the condition.  

I'm not saying this to be rude, but I'm not sure we're looking at the same code.  The code I was commenting on was:
`#include "PluggableUSBHID.h"
#include "USBKeyboard.h"

USBKeyboard Keyboard;

void setup() {
  // put your setup code here, to run once:

}

void loop() {
  // put your main code here, to run repeatedly:
  delay(1000);
  Keyboard.printf("Hello world\n\r");
}`

In that code there is no button control, it is simply an infinitely looping command that uses a function `printf()` which is not defined in the USBKeyboard mbed library anywhere I could find.  When I tried using this example on my board, I could no longer reprogram the board without putting it into bootloader mode (or whatever the proper name is for double pressing reset).  

While I have a whole re-written example using commands from the USBKeyboard mbed library, I was told I should make suggestions that include small changes.  My change was simply to add a way to control whether the example would execute.  I was told this was standard in other keyboard/etc. examples from Arduino, and that this was usually done with a button:
`if button is pressed then print, else do nothing`

alessandromrc:
Hi @Epipeppy no worries :) The code I was actually looking was the one you changed in the PR which is this one:

```cpp
#include "PluggableUSBHID.h"
#include "USBKeyboard.h"

USBKeyboard Keyboard;

// Arbitrary pin.
const int CONTROL_PIN = 2;

void setup() {
  // We will use this to start/end the prints.
  pinMode(CONTROL_PIN, INPUT_PULLUP);
}

void loop() {
  // This will run only if the control pin is connected to ground.
  if( digitalRead(CONTROL_PIN) == LOW){
    delay(1000);
    Keyboard.printf("Hello world\n\r");
  }
}
```
I was related to the line where you have the if statement that checks if the digitalPin is being pulled down. In any way I might look it in a deeper way using a debugger to see if there's something preventing the reset of the uC.

Epipeppy:
Ayy, thank you very much!  It does work for me as well, wonder if it was one of the updates to the mbed core (or rp 2040 library) that fixed the issue I was having back then.  When I ran it just now I did notice that my board disconnects from whatever COM port it was on (I'm on Windows 10 latest update), and is re-detected on a new one.  Either way, I'm glad its not preventing the board from being written to now.

If it's of any interest, I made a few modifications to the USBKeyboard library for my own use.  Because the old keyboard library used `press()`, `release()`, and `send()`, I adapted the `press()` and `sendPressed()` functions to work for this library (added a few other quality of life changes as well).  I would post the request on the correct file, but figured I'd ask first, in case you guys/gals have already been working on updating that.

alessandromrc:
> Ayy, thank you very much!  It does work for me as well, wonder if it was one of the updates to the mbed core (or rp 2040 library) that fixed the issue I was having back then.  When I ran it just now I did notice that my board disconnects from whatever COM port it was on (I'm on Windows 10 latest update), and is re-detected on a new one.  Either way, I'm glad its not preventing the board from being written to now.
> 
> 
> 
> If it's of any interest, I made a few modifications to the USBKeyboard library for my own use.  Because the old keyboard library used `press()`, `release()`, and `send()`, I adapted the `press()` and `sendPressed()` functions to work for this library (added a few other quality of life changes as well).  I would post the request on the correct file, but figured I'd ask first, in case you guys/gals have already been working on updating that.

I've got literally no time to look into the library so if you want you could post your code too ;)

PR #552, opened 2022-09-18
Removed unused variable

alessandromrc:
I removed the char "response" in the method readByte as it was unused and it would just give a compiler warning


facchinm:
Merged, thanks!

Issue #557, opened 2022-09-28
Unable to compile libraries when passing chrono duration instead of integer millisecond count


jvaque:
When compiling any of the examples for USBHID the following warning appears

```
C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\libraries\USBHID\src\USBMouseKeyboard.cpp: In member function 'bool arduino::USBMouseKeyboard::doubleClick()':
C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\libraries\USBHID\src\USBMouseKeyboard.cpp:706:36: warning: 'void rtos::ThisThread::sleep_for(uint32_t)' is deprecated: Pass a chrono duration, not an integer millisecond count. For example use `5s` rather than `5000`. [since mbed-os-6.0.0] [-Wdeprecated-declarations]
     rtos::ThisThread::sleep_for(100);
                                    ^
In file included from C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\libraries\USBHID\src\USBMouseKeyboard.cpp:21:0:
C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\cores\arduino/mbed/rtos/include/rtos/ThisThread.h:216:6: note: declared here
 void sleep_for(uint32_t millisec);
      ^~~~~~~~~
C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\libraries\USBHID\src\USBMouseKeyboard.cpp: In member function 'bool arduino::USBMouseKeyboard::click(uint8_t)':
C:\Users\jvaqu\AppData\Local\Arduino15\packages\arduino\hardware\mbed_nano\3.3.0\libraries\USBHID\src\USBMouseKeyboard.cpp:721:35: warning: 'void rtos::ThisThread::sleep_for(uint32_t)' is deprecated: Pass a chrono duration, not an integer millisecond count. For example use `5s` rather than `5000`. [since mbed-os-6.0.0] [-Wdeprecated-declarations]
     rtos::ThisThread::sleep_for(10);
                                   ^
[...]

PR #558, opened 2022-09-30
Fix compilation warning on sleep_for

jvaque:
On compilation of any project using the USBHID library warnings appear as the library uses a function that was deprecated since mbed 6.0.0.

Previously opened an issue (#557) to get help on the missing imports.


facchinm:
Merged, thank you!

PR #566, opened 2022-10-11
Fix WiFi connection

sallasia:
This fix resolve issue of WiFi connection when in device WiFi range there are many (more then 10) WiFi AP


manchoz:
LGTM. @facchinm @pennam, please review.

facchinm:
LGTM!

Issue #590, opened 2022-12-14
SPI.end() does not check for NULL => crashes NANO33


RobTillaart:
First mentioned here https://github.com/RobTillaart/MS5611_SPI/issues/6

## Summary
It should be safe to call SPI.end() before SPI.begin() to allow dynamic change SPI settings.
Now it crashes on the NANO33

#### Current code
From - https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/SPI/SPI.cpp
```cpp

void arduino::MbedSPI::begin() {
    if (dev == NULL) {
      dev = new mbed_spi;
      dev->obj = NULL;
    }
    if (dev->obj == NULL) {
      dev->obj = new mbed::SPI(_mosi, _miso, _sck);
    }
}

void arduino::MbedSPI::end() {
    if (dev->obj != NULL) {                      //   <<<<<<<<<<  if dev == NULL this crashes.
        delete dev->obj;
    }
}
```

#### Proposal
```cpp
void arduino::MbedSPI::end() {
    if (dev != NULL) {
        if (dev->obj != NULL) {
            delete dev->obj;
        }
    }
}
```

arotman:
This is an important bug fix, as the Arduino code dereferences a NULL pointer and just silently crashes. Ugly stuff. But easy to fix.

PR #594, opened 2022-12-19
SPI: fix crash on end() called before begin()

facchinm:
Fixes https://github.com/arduino/ArduinoCore-mbed/issues/590

@RobTillaart


RobTillaart:
Thanks for fixing so fast Martino!

PR #598, opened 2022-12-23
Change maximum UDP datagram buffer size

aqw42:
UDP packet will not send if its size is greater than 508, and the issue is hard to debug. I made this modification on my side and it works fine. Maybe add a warning if it is exceeded, instead of failing silently ?

### DIFF
--- a/libraries/SocketWrapper/src/MbedUdp.h
+++ b/libraries/SocketWrapper/src/MbedUdp.h
@@ -28,7 +28,7 @@
 #include "netsocket/UDPSocket.h"
 
 #ifndef WIFI_UDP_BUFFER_SIZE
-#define WIFI_UDP_BUFFER_SIZE 508
+#define WIFI_UDP_BUFFER_SIZE 2048
 #endif
 
 namespace arduino {


sebromero:
@aqw42 Hi! I'm not sure I understand. You say "it works fine". Does that mean if you change the packet size it is sent successfully? It's a hardcoded value, so the user is not really supposed to change it. Can you elaborate?

aqw42:
> @aqw42 Hi! I'm not sure I understand. You say "it works fine". Does that mean if you change the packet size it is sent successfully? It's a hardcoded value, so the user is not really supposed to change it. Can you elaborate?

So, if you try to send anything biggest than this value (which is low for a udp datagram) the datagram will be truncated and there is no kind of warning or any way to let the developer know.
I spent some time to find out it was coming  from this hardcoded value. I propose as a temporary solution to avoid such problems    to raise it to a reasonable value (or even a high value, it is the responsibility of the user to not send too big datagrams)

I think a better solution would be implementing a mechanism of parametrisation of this at compile time, plus a way of preventing silent truncation of the datagram (warning at compile time ? Don't know if we can determine statically the size passed to the call) 
As you may have guessed, I don't know how to do this, and my knowledge of this kernel is very limited. 
If you can suggest me a proper way of doing this, I'll be glad to do it
And as a temporary solution I propose this very simple patch.

sebromero:
@aqw42 Got it. Thank you for clarifying. We'll look into it.

Issue #622, opened 2023-02-09
Wire.end() makes the rp2040 crash or hang somewhere, the reason being slave_th.terminate();


TheWorkshopByGM:
I am trying to end the I2C to start a new one with a new address (for software addressing purposes). But every time I try to use Wire.end() it just hangs or crashes.

facchinm:
Hi @TheWorkshopByGM ,
which core version are you using? The patch https://github.com/arduino/ArduinoCore-mbed/commit/b441a19058a3f3f0f157d21cef782e9f4cacbb09 is included in 3.5.4 and in fact doesn't set `slave` to `NULL` when calling `end()`, which could create issues if you switch role.
I'll post a PR soon to target the issue, if you could test it it would be great

PR #623, opened 2023-02-09
Wire: nullify pointers on end()

facchinm:
Should fix https://github.com/arduino/ArduinoCore-mbed/issues/622

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -45,6 +45,7 @@ void arduino::MbedI2C::begin(uint8_t slaveAddr) {
 void arduino::MbedI2C::end() {
 	if (master != NULL) {
 		delete master;
+		master = NULL;
 	}
 #ifdef DEVICE_I2CSLAVE
 	if (slave != NULL) {
@@ -52,6 +53,7 @@ void arduino::MbedI2C::end() {
 		slave_th->free_stack();
 		delete slave_th;
 		delete slave;
+		slave = NULL;
 	}
 #endif
 }


TheWorkshopByGM:
Just tried it, got a few errors. I think it should be:
if (slave != NULL) {
    slave_th.terminate();
    slave_th.free_stack();
    delete slave;
    slave = NULL;
}

But it still crash.  If I remove the slave_th.terminate(). It stops crashing end ends the I2C slave. But if I try to start the I2C again using Wire.begin(new_address); it keeps the old address.

For example this is the code I am running on the Master:
`#include <myWire.h>

const int newAddress = 8; //new address i want to switch to
bool addressChanged = false; //bool to store if add has been changed

const int changeAddCom = 1;

void setup() {
  Wire.begin();
  Serial.begin(9600);
  pinMode(A0, INPUT);
}

void loop() {
  if(addressChanged == false){
    Serial.println("Will attempt to change address!");
    Wire.beginTransmission(0); // Address 0 is the default address of the slave
    Wire.write(changeaAddCom); // Send the command to change the address to the slave
    Wire.write(newAddress); // Send the new address to the slave
    Wire.endTransmission();
    delay(100);
    addressChanged = true;
  }
  else{
    Serial.println("Testing which address is working"); // will send the message 0 to add 0 and 8 to add 8, and will see what gets printed on the slave
    Wire.beginTransmission(0);
    Wire.write(0); 
    Wire.endTransmission();
    Wire.beginTransmission(8); 
    Wire.write(8); 
    Wire.endTransmission();
    delay(250);
  }
}`

This is the code for the slave:
`#include <myWire.h>

const int ledPin = 29;
bool addressUpdated = false;

void setup() {
  Wire.begin(0);
  Wire.onReceive(receiveEvent);
  Serial.begin(9600);
  pinMode(ledPin, OUTPUT);
}

void loop() {
  delay(100);
}

void receiveEvent(int bytesReceived) {
  Serial.println("message received!");
  int command = int(Wire.read());
  Serial.print("Command: ");
  Serial.print(command);
  if(command == 1){
    Serial.println("Command = 1!");
    int data = int(Wire.read());
    Serial.print(" - Data: ");
    Serial.println(data);
    if(addressUpdated == false){
      Serial.println("Will end!");
      Wire.end();
      Serial.println("Ended!");
      delay(500);
      Serial.println("Begin");
      Wire.begin(8); // Change address
      Serial.print("Address changed to: ");
      Serial.println(data);
      addressUpdated = true;
    }
  }
  else{
    Serial.print("I2C Received - ");
    Serial.print("Message: ");
    Serial.println(command);
  }
}`

facchinm:
@TheWorkshopByGM it appears you are applying on a different version of the core. Make sure you update to 3.5.4 before doing any other test

PR #628, opened 2023-02-20
Updated naming of products

jhansson-ard:
Updated some products to their proper name

@facchinm

### DIFF
--- a/extras/package_index.json.NewTag.template
+++ b/extras/package_index.json.NewTag.template
@@ -335,7 +335,7 @@
           },
           "boards": [
             {
-              "name": "Arduino Opta"
+              "name": "Opta™"
             }
           ],
           "toolsDependencies": [
--- a/extras/package_index.json.NewTag.template
+++ b/extras/package_index.json.NewTag.template
@@ -42,7 +42,7 @@
               "name": "Raspberry Pi Pico"
             },
             {
-              "name": "Nicla Sense ME"
+              "name": "Arduino Nicla Sense ME"
             },
             {
               "name": "Arduino Nicla Vision"
@@ -284,13 +284,13 @@
           },
           "boards": [
             {
-              "name": "Nicla Sense ME"
+              "name": "Arduino Nicla Sense ME"
             },
             {
-              "name": "Nicla Voice"
+              "name": "Arduino Nicla Voice"
             },
             {
-              "name": "Nicla Vision"
+              "name": "Arduino Nicla Vision"
             }
           ],
           "toolsDependencies": [
@@ -335,7 +335,7 @@
           },
           "boards": [
             {
-              "name": "Arduino Finder Opta"
+              "name": "Arduino Opta"
             }
           ],
           "toolsDependencies": [


extras/package_index.json.NewTag.template
```diff
@@ -335,7 +335,7 @@
           },
           "boards": [
             {
-              "name": "Arduino Finder Opta"
```
facchinm, line 338: I'd ping Francesca before changing this one

extras/package_index.json.NewTag.template
```diff
@@ -335,7 +335,7 @@
...
+              "name": "Arduino Opta"
```
jhansson-ard, line 338:
```suggestion
              "name": "Opta®"
```

jhansson-ard: @facchinm according to Francesca, this should be only "Opta". Can we also use the ® in here or will stuff break?

extras/package_index.json.NewTag.template
jhansson-ard: see comment below!

extras/package_index.json.NewTag.template
facchinm: It could be ok for me but in the landing page it's clearly called "Arduino Opta" https://store.arduino.cc/pages/opta, so we should try to be at least consistent. ® shouldn't be a problem but let's not add it everywhere or it becomes a mess

jhansson-ard: For all the places we share with Finder, the naming should be Opta® and the places we have for ourselves (the store page for example) uses the Arduino Opta™. This is true for at least documentation for the docs tutorials and datasheet ⭐

jhansson-ard: Here's more on the subject: https://arduino.atlassian.net/wiki/spaces/HWFPLC/pages/3627450402/Trademark+use

jhansson-ard: Oh sorry, it should be Opta™ here!

extras/package_index.json.NewTag.template
```diff
@@ -335,7 +335,7 @@
           },
           "boards": [
             {
-              "name": "Arduino Finder Opta"
+              "name": "Opta™"
```
sebromero, line 338:
```suggestion
              "name": "Arduino Opta"
```
Do we really want to use trademark symbols here? What about the other boards?

sebromero (changes requested):
See comment

extras/package_index.json.NewTag.template
jhansson-ard: This is the right way to write the name here:

"It's Arduino Opta® (our own documentation) or Opta™ (shared documentation)" [last comment here](https://github.com/arduino/ArduinoCore-mbed/pull/628#discussion_r1112104729)

extras/package_index.json.NewTag.template
jhansson-ard: Here: https://github.com/arduino/ArduinoCore-mbed/pull/628#discussion_r1112104729

extras/package_index.json.NewTag.template
jhansson-ard: According to Francesca Gentile, this documentation is shared with Finder - and that is why we cannot use the `Arduino Opta` naming.

PR #633, opened 2023-03-06
Provide CAN library for enabling useage of CAN for ArduinoCore-mbed enabled boards.

aentinger:
**Caveat**: which also make CAN available on their GPIOs.

Requires: https://github.com/arduino/ArduinoCore-API/pull/185 .


variants/PORTENTA_H7_M7/pins_arduino.h
```diff
@@ -164,4 +164,12 @@ void _ontouch1200bps_();
 
 #define USB_MAX_POWER	(500)
 
+#define CAN_HOWMANY       2
+
+#define PIN_CAN0_TX       (PB_13)
+#define PIN_CAN0_RX       (PB_5)
+
+#define PIN_CAN1_TX       (PH_13)
```
facchinm, line 172: On Portenta we only expose one CAN bus in the HD connectors, we should probably expose only that one to the user (PH13/PB8)
2023-04-26-123445_922x44_scrot

facchinm (changes requested):

variants/PORTENTA_H7_M7/pins_arduino.h
aentinger: Sorry, I must have overlooked the notification for your comment. I've fixed this in f4cad9f119ed6a18cd8211fecac254347debcde0 .

**Note**: I'm not really happy about the divergence between software object (CAN) and labeling on the high-density connector (CAN1) but if we follow the usual Arduino policy of enumerating software objects by amount of available peripherals (i.e. the first CAN peripheral is mapped to CAN, the second CAN peripheral maps to CAN1, etc.) then in this case mapping the bus labelled CAN1 to a software object called CAN is only a logical (if unfortunate) choice.

What are your thoughts on this?

facchinm (approved):

variants/PORTENTA_H7_M7/pins_arduino.h
aentinger: Alright, good for me :wink: . I've noted this discrepancy in the comments, at least that should help if/when someone falls over this.

PR #635, opened 2023-03-08
Added return to function enterShipMode

alessandromrc:
I've added the code that actually checks if the value has been written and then returns its value to resolve this issue: "warning: no return statement in function returning non-void".

### DIFF
--- a/libraries/Nicla_System/src/Nicla_System.cpp
+++ b/libraries/Nicla_System/src/Nicla_System.cpp
@@ -100,6 +100,10 @@ bool nicla::enterShipMode()
   uint8_t status_reg = _pmic.getStatus();
   status_reg |= 0x20;
   _pmic.writeByte(BQ25120A_ADDRESS, BQ25120A_STATUS, status_reg);
+  if (_pmic.readByte(BQ25120A_ADDRESS, BQ25120A_STATUS) != status_reg) {
+    return false;
+  }
+  return true;
 }
 
 uint8_t nicla::readLDOreg()


sebromero:
Thank you for your pull request. While this works, I think I would prefer to fix this in the PMIC driver code that should return the success / failure status:

```cpp
bool BQ25120A::writeByte(uint8_t address, uint8_t subAddress, uint8_t data)
{
  cd = 1;
  nicla::i2c_mutex.lock();
  Wire1.beginTransmission(address);
  Wire1.write(subAddress);
  Wire1.write(data);
  uint8_t result = Wire1.endTransmission();
  nicla::i2c_mutex.unlock();
  cd = 0;
  return result == 0;
}
```

Then in the enterShipMode function could just piggy back on that return status:

```cpp
bool nicla::enterShipMode()
{
  //    STATUS reg:
  //    | B7 | B6 |      B5     | B4 | B3 | B2 | B1 | B0 |
  //    | RO | RO | EN_SHIPMODE | RO | RO | RO | RO | RO |

  uint8_t status_reg = _pmic.getStatus();
  status_reg |= 0x20;
  return _pmic.writeByte(BQ25120A_ADDRESS, BQ25120A_STATUS, status_reg);
}
```

@facchinm Thoughts?

facchinm:
`enterShipMode()` should immediately "kill" the board if I remember correctly, so the return is not going to be taken into consideration (except to silence the compiler :slightly_smiling_face: )

alessandromrc:
@facchinm I would add it in every case, firstly to silence the compiler like you said but we actually don't know what might be going to happen on the microcontroller that might cause issues.

PR #647, opened 2023-03-23
Make WiFiClientSSL class available

karlsoderby:
`WiFiSSLClient` was missing in `WiFi.h`, so examples using SSL fails to compile.

The SSL example was tested with the GIGA R1 WiFi board successfully.

### DIFF
--- a/libraries/WiFi/src/WiFi.h
+++ b/libraries/WiFi/src/WiFi.h
@@ -197,6 +197,7 @@ public:
   friend class WiFiClient;
   friend class WiFiServer;
   friend class WiFiUDP;
+  friend class WiFiSSLClient;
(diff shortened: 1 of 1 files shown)


sebromero (approved):
LGTM

sebromero:
@karlsoderby Where is the example sketch that you're referring to?

facchinm (changes requested):
Take into account the flash occupation increase

sebromero:
@karlsoderby Can you please point me to the examples that were not compiling?

manchoz:
Hi,
Can't we just include `WiFiSSLClient.h` header in the SSL/TLS examples?

PR #656, opened 2023-04-04
[AE-39] Add API documentation to Camera library

sebromero:
This PR adds the missing documentation to the Camera library. Currently the API documentation is rendered by hand. This process will be automated with a future PR.


libraries/Camera/docs/README.md
alrvid, line 1: Should the Himax HM01B0 be mentioned too in the list of compatible camera models?

libraries/Camera/library.properties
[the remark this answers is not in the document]
alrvid: Should the fourth camera model be mentioned here too?

libraries/Camera/src/camera.h
alrvid, line 1: Should the fourth camera model be mentioned?

alrvid: Misspelling on line 87.

alrvid: The reset() function isn't specifically for getting out of stuck states. What it does differs between the camera models. It does nothing at all on the GC2145. On the OV7670, it reset all configuration registers to the default values. The registers are reset to the defaults on the Himax cameras too, but they also enter Standby Mode, where there is no video output and the power consumption is lowered.

libraries/Camera/src/camera.h
```diff
@@ -14,83 +14,315 @@
...
+
+        /**
+         * @brief Output debug information to a stream.
+         * You can use this function to output debug information to the serial port by passing Serial as the stream.
+         * @param stream Stream to output the debug information
+         */
         virtual void debug(Stream &stream) = 0;
 
+        /**
+         * @brief Set the sensor in standby mode.
```
alrvid, line 286: This actually has no effect on none of the currently supported cameras. The Himax cameras can be set in standby mode by calling reset() instead, but not by calling this function.

alrvid (changes requested):
See the comments for each file.

libraries/Camera/docs/README.md
sebromero: Yes

PR #657, opened 2023-04-07
nicla: Add missing return value in enterShipMode function.

sebromero:
An alternative solution to #635 which adds a missing return statement that leads to a compiler warning.

Thank you @alessandromrc for spotting this!


alrvid (approved):
This looks good to me :)

PR #664, opened 2023-04-20
Update wiring_analog.cpp

alessandromrc:
Fixes a warning about comparing an integer to an unsigned integer

### DIFF
--- a/cores/arduino/wiring_analog.cpp
+++ b/cores/arduino/wiring_analog.cpp
@@ -61,7 +61,7 @@ void analogWrite(pin_size_t pin, int val)
     return;
   }
 #ifdef DAC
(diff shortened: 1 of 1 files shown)


facchinm:
Merged, thanks for your contribution!

alessandromrc:
@facchinm I wanted to ask you if there's a possibility to avoid the Nicla Voice from always failing the test as it can mislead people when they see a board failing to compile

facchinm:
Sure, I'm trying to understand which revision of arduino-audio-tools library introduced the regression (and pin to a previous version)

facchinm:
Should be fixed by https://github.com/arduino/ArduinoCore-mbed/pull/668

PR #667, opened 2023-04-22
[Camera]  Fix a synchronization bug in example

alrvid:
Fixes a synchronization bug. Return if there's no timeout OR if there's no sync byte received. The previous version needed both to be true at the same time to return without sending a frame, which meant that a frame was sent if there was a timeout even if no sync byte had been received. That, in turn, caused random problems for the Processing sketch on the other end.


alrvid:
The new functionality uses the timeouts to set the (constant) frame rate. If timeouts are needed for lost sync bytes, they should be implemented on the other end to keep the synchronization robust.

PR #668, opened 2023-04-26
Nicla_voice: fix gh action (compile examples)

facchinm:


alessandromrc:
Let's hope this works now :)

PR #670, opened 2023-05-02
Nicla voice: Recorder example correct fix

facchinm:


sebromero:
@facchinm Records random noise with my board. Please check the attached files.
[sound.zip](https://github.com/arduino/ArduinoCore-mbed/files/11372750/sound.zip)

mbmn974:
> @facchinm Records random noise with my board. Please check the attached files. [sound.zip](https://github.com/arduino/ArduinoCore-mbed/files/11372750/sound.zip)

Hello, sorry for using this space to ask my question but I don't find any other appropriate places.

How did you manage to record the sound of the nicla voice ? Did you use the example "Record_and_stream" ? I believe that it isn't enough ?

Thanks in advance.

sebromero:
Hi @mbmn974 👋  No worries. When running that example you need to extract the raw data from the audio stream and save it to your computer. The instructions for this are in the comments in the header of the example sketch. Please let me know if your recording sounds okay or if you also get noise like I did.

mbmn974:
Hi, thanks for your quick response.
I managed to record the audio from the nicla voice's microphone using the header's instructions.
The audio has a constant noise but it also detects other sounds that I made, like discussions or keyboards sounds. I'm currently in a noisy environment with ventilation, maybe the noise comes from it. I will test it back home, if there is no more background noise I'll update you.

mbmn974:
Update : after testing it in a quiet environment there is no background noise.

Issue #632, opened 2023-03-03
Arduino MBED Version 4.X bootloader Question.


hpssjellis:
How important is it to load the MBED new bootloader for the Portenta? Mine is presently at Bootloader version: 23

What is the default question and answers that you would expect for developers? I get that final products would use the secure boot. 

1.  "Do you want to install/update the default Arduino bootloader? Y/[n]
Choosing "No", will install/update the MCUboot bootloader."  ?

Too scared to go farther. Looks like either choice does something to the board.

Is there a process to reverse the secure boot if you mess-up (I haven't yet, just asking). I assume there is no way to reverse a bad secure boot which would basically brick the board.

pennam:
Hi @hpssjellis,
The bootloader update `v24` introduced with mbed core release `4.0.2` is only for the Nicla Vision. The bootloader v24 for the portenta was introduced with release `3.4.1`.

Difference between version `v23` and `v24` is the way bootloader handles RTC clock at boot in order to read RTC backup registers: with newer cores > `3.4.1` and old bootloaders < `v24` there are some possible conditions in which backup registers are not retained with a board reset and OTA will not work.

For Portenta H7 the [STM32H747_manageBootloader](https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/STM32H747_System/examples/STM32H747_manageBootloader/STM32H747_manageBootloader.ino) sketch offers the user the possibility to flash [mcuboot-arduino-stm32h7](https://github.com/arduino/mcuboot-arduino-stm32h7) bootloader that offers the possibility to enable and evaluate some security features included in the [mcuboot](https://github.com/mcu-tools/mcuboot) bootloader.

The `mcuboot-arduino-stm32h7` can be used in two modes:
* Flashing only the bootloader but not the [keys](https://github.com/arduino/ArduinoCore-mbed/blob/64d42ffb0021352dd4a8acd53b461d8466523413/libraries/STM32H747_System/examples/STM32H747_manageBootloader/STM32H747_manageBootloader.ino#L157) will make the bootloader work in the same exact way the "standard" bootloader is working, no changes and no security features enabled. 

[...]

PR #672, opened 2023-05-02
Bootloader v25 update proposal

pennam:
Since form the next portenta batch the default bootloader version will be v25 and it will include by default MCUboot security features (as for OPTA and GIGA). My proposal is to split the bootloader management from the security feature setup because:

* key management can scare users who just want to update the bootloader https://github.com/arduino/ArduinoCore-mbed/issues/632
* sketches will be easier to maintain

/cc @marqdevx


libraries/MCUboot/examples/enableSecurity/enableSecurity.ino
```diff
@@ -0,0 +1,155 @@
...
+}
+
+void applyUpdate() {
+  flash.init();
+  setupMCUBootOTAData();
+  flash.program(&enc_priv_key, ENCRYPT_KEY_ADDR, ENCRYPT_KEY_SIZE);
+  flash.program(&ecdsa_pub_key, SIGNING_KEY_ADDR, SIGNING_KEY_SIZE);
+  Serial.println("Flashed 100%");
+  flash.deinit();
+  Serial.println("\nBootloader update complete. It's now safe to reboot or disconnect your board.");
```
sebromero, line 150: This is no longer a bootloader update, so we should probably rephrase this.

libraries/MCUboot/examples/enableSecurity/enableSecurity.ino
```diff
@@ -0,0 +1,155 @@
+#include "FlashIAP.h"
```
sebromero, line 1: Can we take the opportunity to add a comment section here that explains the purpose of the sketch, what it does and how it works?

sebromero (changes requested):
Thanks for the PR, good move! I haven't tested it but I agree with the general approach of splitting the two. Maybe we can find a name for the sketch that explains a bit better what it does? It's about installing the keys to enable encryption, right?
Please also have a look at my other comments.

libraries/MCUboot/examples/enableSecurity/enableSecurity.ino
pennam: yup :100: i had pushed the change in another branch

libraries/MCUboot/examples/enableSecurity/enableSecurity.ino
pennam: done !

pennam:
> Maybe we can find a name for the sketch that explains a bit better what it does? It's about installing the keys to enable encryption, right?

The original name of this sketch was writeKeys.ino. This is exactly what is does, it writes two keys in the flash memory:
* one is used for the image signature checking
* and the other is used for image encryption

writing the key will enable MCUboot image signature checking and image swap with encryption/decryption so i decided to rename it to enableSecurity, but i agree it is not the best, any suggestion?

PR #680, opened 2023-05-04
[AE-74] Revamp Battery API for Nicla Sense ME

sebromero:
# Change log

## 🐛 Fixed Bugs
- Fixed the incorrect calculation of the battery voltage percentage.
- Fixed incorrect voltage reading by adding a tiny delay to give the chip time to wake up from High Impedance mode.
- Fixed a slightly incorrect interpretation of the NTC values (introduced a new state 'Extreme' which indicates the charging has been suspended).
- Fix edge case in the charging current setting that could cause an external value to be used instead of a programmatic one.
- Fix incorrect TS and interrupt disable in enableCharging function

## 🚀 Enhancements
- Added documentation to all functions related to power management.
- Added error handling to catch error cases
- Introduced Enums for battery charge level, temperature and operating status.
- Avoided high impedance mode when connected through VIN to to correctly report charging status
- Improved accuracy of battery voltage reading (disable charging while reading)

## 🆕 New Features
- Implemented a WebBLE based battery monitor + sketch for easy testing without external power supply
- Remove battery status function and replace with temp and charge level.
- Added function for toggling battery NTC.
- Added API for operating system status (e.g. charging).
- Added API to set regulated battery voltage.
- Added function to determine power source
- Added function to disable charging state


aliphys:
LATM!

😃 Nicla - PC WebBLE communication works as described. 
😃 Connection is stable, even after several hours
😃 Battery graphic is cute / informative 

😔 Reported battery voltage differs between WebBLE interface (3.86V) and scope 

--- 
- Webpage running locally on Chrome in Ubuntu 22.04. Voltage reported as 3.86V
    - image
- Voltage between VBAT pin and USB ground is 3.32V
    - img_20230504_180717_720
    - NIC-BAT

---

This difference could be due to internal resistance of the ground plane.

libraries/Nicla_System/extras/BatteryMonitor/index.html
```diff
@@ -0,0 +1,38 @@
...
+    <script defer src="app.js"></script>
+</head>
+<body>
+    <h1>WebBLE Battery Monitor 🔋</h1>
+    <div class="container">
+        <div class="battery">
+            <div class="battery-level" id="battery-level"></div>
+        </div>
+        <p id="battery-label"></p>
+        <div id="battery-status">
```
aliphys, line 22: The two icons are not displayed in my system.

aliphys: image

sebromero: @aliphys They should only be displayed when you power the board through USB and charge the battery respectively. Can you try?

sebromero:
@aliphys Odd, I get a pretty accurate reading. The internal resistance of the ground plane should be neglectable.
Screenshot 2023-05-05 at 09 36 34
IMG_5019

sebromero:
@aliphys Found another bug and fixed it. Basically the battery reading was not done correctly so it would only ever use what is measured when you plug it in.

sebromero:
@aliphys Found another bug and fixed it. Basically the battery reading was not done correctly so it would only ever use what is measured when you plug it in.

sebromero:
@aliphys Found another bug and fixed it. Basically the battery reading was not done correctly so it would only ever use what is measured when you plug it in.

sebromero:
@facchinm
Some observations:
- 📖 A write operation was gonna be used to kick the watchdog previously, however the way it was implemented didn't seem to trigger a write operation. **Conclusion: A read operation seems to be enough to reset the watchdog timer.** 
- ⚡ Battery seems to charge even when NTC settings are enabled but the battery doesn't have one. **Conclusion: Accidentally enabling NTC on a non-NTC battery doesn't prevent it from being charged.**
- 🔋 If the battery doesn't have an NTC, the speed is automatically reduced to ~35mA no matter the setting. This one may need some more investigation.
- 🐶 The Watchdog mechanism seems to be redundant now after reworking the library:
  - The watchdog is disabled in High-Z mode. Now, by default the PMIC doesn't ever leave High-Z mode when powered through VIN (CD only used to toggle the charging). When powered with a battery it only leaves it very shortly to read some bytes from the PMIC within the 50s timeout of the watchdog. The new `disableCharging()` function was implemented by modifying a register, not by using CD. **Conclusion: We could probably delete all code related to the watchdog timer.**

That aside, I think I'm done here. Waiting for @aliphys ' green light 🟢

aliphys (approved):
LGTM. Thanks for the detailed PR and comments: we should try to strive for this level of quality.
Battery voltage updates correctly 👍

libraries/Nicla_System/examples/NiclaSenseME_BatteryStatus/NiclaSenseME_BatteryStatus.ino
```diff
@@ -0,0 +1,338 @@
...
+ * The BatteryMonitor web app can also be configured to poll the battery status every X seconds.
+ * 
+ * The LED colors are used to indicate the BLE connection status:
+ * - Green: Board is ready but no BLE connection has been established yet.
+ * - Blue: Board is connected over BLE to a device that wants to read the battery status.
+ * - Red: A device that was previously connected over BLE has disconnected.
+ * 
+ * Instructions:
+ * 1. Upload this sketch to your Nicla Sense ME board.
+ * 2. Open the BatteryMonitor web app (index.html) in a browser that supports Web Bluetooth (Chrome, Edge, Opera, etc.).
```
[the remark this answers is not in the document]
sebromero: @aliphys No longer required for Chrome. It's still experimental but enabled by default.

Issue #87, opened 2020-11-09
mbed-os-to-arduino with MacOS


sabas1080:
Hi

I am trying to compile my own variant, but when executing the command:

`./mbed-os-to-arduino -r /Users/xxxx/Documents/xxxx/mbed-os BASTBLE:BASTBLE`

I get the following error

```
./mbed-os-to-arduino: line 273: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

facchinm:
Hi @sabas1080 ,
trying to make the script correct for osx is a bit of a nightmare. One of the open PRs here https://github.com/arduino/ArduinoCore-nRF528x-mbedos/pulls could firx your problem but it's not a long term solution (and OSX support of unix utilities changes every release).
We are currently rewriting it in python for full cross platform compatibility :slightly_smiling_face: 
In the meantime I'd say that Linux is the only compatible platform

PR #694, opened 2023-06-08
Make README clear about mbed-to-arduino script compatability

trylaarsdam:
It would be nice if the readme specified that the `mbed-os-to-arduino` script is not compatible with MacOS, and has to be run on a Linux system.

According to @facchinm in #87 the script was not compatible as of 2020, and after trying it out on MacOS today it is still not working because of the `declare -A` command being different in MacOS than on Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ The repository contains the Arduino APIs and IDE integration files targeting a g
 
 ## Installation
 
+Note: 
+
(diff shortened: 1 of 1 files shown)


facchinm:
Thanks @trylaarsdam , we tried a few times to port the script to macOS but without much success, so the clarification is perfect.

PR #698, opened 2023-06-16
Add guard for pin in range in pinMode

### LINKED ISSUE #693: pinMode does not accept NC / unused pin
Normally, if a library takes a pin that is not used, I would pass in `-1` as the pin number. The equivalent in Mbed would be `NC`. However, this `-1` crashes the MCU with the mbed core. (I'm not sure if crashes is the right term. My Portenta H7 is not responding and the red LED is blinking).

The reason is that libraries made for arduino use [`pinMode` with this signature](https://github.com/arduino/ArduinoCore-mbed/blob/main/cores/arduino/wiring_digital.cpp#L52):
```cpp
void pinMode(pin_size_t pin, PinMode mode)
```

There is also a version with `PinName`, but most arduino libraries use `uint8_t` or something like that for their pin numbers, not `PinName`. 

[This definition of `pinMode`](https://github.com/arduino/ArduinoCore-mbed/blob/main/cores/arduino/wiring_digital.cpp#L52) looks up the GPIO object inside of `g_APinDescription` without doing any checks that the pin is in range.

The equivalent definitions of `digitalRead` and `digitalWrite` check if the pin is in range before doing anything, but `pinMode` does not.

```cpp
  if (pin >= PINS_COUNT) {
    return LOW;
  }  
```

I think `pinMode` should have the exact same guard as `digitalRead` and `digitalWrite`. I can open a PR for this. Is there anything I am missing as to why this is implemented this way?

MarcelRobitaille:
Fix #693

The [`pinMode(pin_size_t pin, PinMode mode)`](https://github.com/arduino/ArduinoCore-mbed/blob/main/cores/arduino/wiring_digital.cpp#L52) function can be passed a pin number that is out of range, crashing the microcontroller. This PR adds the same test found in [`digitalWrite(pin_size_t pin, PinStatus val)`](https://github.com/arduino/ArduinoCore-mbed/blob/main/cores/arduino/wiring_digital.cpp#L91) and [`digitalRead(pin_size_t pin)`](https://github.com/arduino/ArduinoCore-mbed/blob/main/cores/arduino/wiring_digital.cpp#L114) to ensure that the pin is in range.

### DIFF
--- a/cores/arduino/wiring_digital.cpp
+++ b/cores/arduino/wiring_digital.cpp
@@ -51,6 +51,9 @@ void pinMode(PinName pin, PinMode mode)
 
 void pinMode(pin_size_t pin, PinMode mode)
 {
+  if (pin >= PINS_COUNT) {
+    return;
+  }
   mbed::DigitalInOut* gpio = digitalPinToGpio(pin);
   if (gpio != NULL) {
     delete gpio;


MarcelRobitaille:
@facchinm Could you take a look?

facchinm:
Thank you for the ping @MarcelRobitaille , I lost the notification apparently :slightly_smiling_face:

PR #701, opened 2023-06-28
[AE-124] [PROBUPOC-694] Enable D15-D18 to reference digital pins on the Opta

aliphys:
Referencing pins D15 to D18 on the Opta leads to a `Pin cannot be used as digital pin` error. This is traced to the following lines in [pins_arduino.h](https://github.com/arduino/ArduinoCore-mbed/blob/main/variants/OPTA/pins_arduino.h).
https://github.com/arduino/ArduinoCore-mbed/blob/aef18a14398de1a7871d49a7de63fcb5d3643818/variants/OPTA/pins_arduino.h#L88-L91

This PR allows for D15, D16, D17 and D18 to be used within Arduino sketches on the Opta by defining them accordingly. 
image

### DIFF
--- a/variants/OPTA/pins_arduino.h
+++ b/variants/OPTA/pins_arduino.h
@@ -85,10 +85,10 @@ static const uint8_t A7  = PIN_A7;
 #define D12 (12u)
 #define D13 (13u)
 #define D14 (14u)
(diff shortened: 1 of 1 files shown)


manchoz:
LGTM

PR #711, opened 2023-08-17
post_install.sh: Use uaccess tag in udev rules

tormodvolden:
Do not make device node world-writeable, instead give access to desktop user.

Also use one "here" document instead of mixing in echo statements. Customarily keep "then" on same line as "if".

### DIFF
--- a/post_install.sh
+++ b/post_install.sh
@@ -1,19 +1,19 @@
 #!/usr/bin/env bash
 
 arduino_mbed_rules () {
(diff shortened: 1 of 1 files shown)


tormodvolden:
(fixed up commit message)

tormodvolden:
Signed the CLA with the help of this: https://github.com/cla-assistant/cla-assistant/issues/730#issuecomment-913610724

PR #734, opened 2023-10-10
mbed-os-to-arduino macOS script

leonardocavagnis:
mbed-os-to-arduino script to compile a blank Mbed OS project on macOS system

FYI @manchoz


AndrewCapon:
Thanks, This is very useful.

Will give this a look today...

PR #749, opened 2023-10-31
Add macAddress() getter

alranel:
This PR adds a new signature for macAddress:

```
String macAddress();
```

This signature is already present in ESP32, so adding it to this core will increase the interoperability with sketches and libraries developed for ESP32.


facchinm:
@alranel thanks for the PR, can you also provide one of Ethernet library?

JAndrassy:
> can you also provide one of Ethernet library?

both WiFiClass and EthernetClass inherit from MbedSocketClass

facchinm:
> both WiFiClass and EthernetClass inherit from MbedSocketClass
right, I slipped over while reading the commit message, merging

PR #750, opened 2023-10-31
Add WiFiServer::accept() as alias to WiFiServer::available()

alranel:
Since `available()` is currently implemented as `accept()` (i.e. it returns a client even before it has actually sent any data), we provide an alias for it.


JAndrassy:
maybe EthernetServer could get accept() in this PR too.

pennam:
superseded by https://github.com/arduino/ArduinoCore-mbed/pull/793

PR #751, opened 2023-10-31
Add WiFiServer::stop() and WiFiServer::end()

alranel:
This PR adds an `end()` method (aliased to `stop()` for better interoperability) which is symmetric to `begin()`. This is borrowed by the ESP8266 and ESP32 cores.


pennam:
superseded by https://github.com/arduino/ArduinoCore-mbed/pull/793

PR #755, opened 2023-11-07
Improve reliability of the Callback implementation

alrvid:
These changes are made since some combinations of code and compiler versions sometimes lead to erroneous machine code being generated, because of the somewhat adventurous implementation in Callback.h. The first change prevents the GCC compiler from applying optimizations that assume the code follows strict aliasing rules. The second one prevents the GCC compiler from sometimes generating incorrect inline code for the Callback constructor. The machine code generated with these changes are, on the one hand, less optimal. But on the other hand, the changes should prevent bugs arising from undefined behavior that is tricky to find in the Callback implementation, or simply from compiler bugs in GCC. The philosophy here is that it's better to be safe than sorry, rather than trying to find and fix every single dangerous construct in the code.

See also this post from a few years ago by one of the authors of Callback.h: https://answers.launchpad.net/gcc-arm-embedded/+question/686870

### DIFF
--- a/cores/arduino/mbed/platform/include/platform/Callback.h
+++ b/cores/arduino/mbed/platform/include/platform/Callback.h
@@ -17,6 +17,11 @@
 #ifndef MBED_CALLBACK_H
 #define MBED_CALLBACK_H
 
+// This prevents the GCC compiler from applying optimizations that assume the code follows strict aliasing rules.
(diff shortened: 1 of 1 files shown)


facchinm:
Ouch, I noticed after merging that the patch will be overridden after the next rebuild of the core... I'll reapply at https://github.com/arduino/mbed-os and get the patch back in.

PR #768, opened 2023-11-21
SocketWrapper - copyable networking clients

JAndrassy:
copyable networking clients

it is done the same way as in other LwIP based Arduino networking libraries including the C33 LwipWrapper. There WiFiClient/EthernetClient is a simple copyable class redirecting every method to the object managing the socket. That object is stored in std:shared_ptr.
The PR uses the existing MbedClient as the managing class and adds the copyable class from which then WiFiClent and EthernetClient inherit.

I can't test WiFiSSLSE050Client

### DIFF
--- a/libraries/SE05X/src/WiFiSSLSE050Client.h
+++ b/libraries/SE05X/src/WiFiSSLSE050Client.h
@@ -23,18 +23,17 @@
 
 #include "SE05X.h"
 #include "WiFiSSLClient.h"
+#include "MbedSSLClient.h"
 
 extern const char CA_CERTIFICATES[];
 
 namespace arduino {
 
-class WiFiSSLSE050Client : public arduino::WiFiSSLClient {
+class MbedSSLSE050Client : public arduino::MbedSSLClient {
 
 public:
-  WiFiSSLSE050Client();
-  virtual ~WiFiSSLSE050Client() {
-    stop();
-  }
+  MbedSSLSE050Client();
+
   void setEccSlot(int KeySlot, const byte cert[], int certLen);
 
 private:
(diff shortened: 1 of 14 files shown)


JAndrassy:
TODO adopt WiFiSSLSE050Client

megacct:
Great work, I'm back up and running with my telnet server. I did get a crash when doing a client available() or read() (not sure which) after having called a stop() on the client. Easy workaround in my code and probably shouldn't have been doing that anyway.

JAndrassy:
@facchinm this is important. the networking libraries in this core will not work correctly without this. the copy ability of instanced objects is a feature of the Arduino language. Imagine String would not be copyable in some Core.

pennam:
@JAndrassy I think you should rebase this on master since `setTimeout` has been renamed in `setSocketTimeout`

JAndrassy:
@pennam thank you for the reminder. done

libraries/SE05X/src/WiFiSSLSE050Client.h
```diff
@@ -65,6 +64,14 @@ class WiFiSSLSE050Client : public arduino::WiFiSSLClient {
   }
 };
 
+class WiFiSSLSE050Client : public arduino::WiFiSSLClient {
+
+  void setEccSlot(int KeySlot, const byte cert[], int certLen);
```
pennam, line 69:
```suggestion

public:
  void setEccSlot(int KeySlot, const byte cert[], int certLen);
```

pennam (changes requested):
Thanks for the contribution @JAndrassy 

Do you mind also changing `setEccSlot` visibility? It must be public.

I'm running a bit of testing on this PR so we can hopefully merge it for the next release.

pennam (approved):
:+1:  Thanks

PR #769, opened 2023-11-22
SocketWrapper: avoid Stream setTimeout shadowing

### LINKED ISSUE #728: SocketWrapper library: `MbedClient.setTimeout` shadows `Stream.setTimeout`
`MbedClient` has `setTimeout` instead of `setConnectionTimeout`. `setTimeout` shadows `Stream.setTimeout`

pennam:
Shoould fix #728

### DIFF
--- a/libraries/SocketWrapper/src/MbedClient.h
+++ b/libraries/SocketWrapper/src/MbedClient.h
@@ -97,7 +97,7 @@ public:
   IPAddress remoteIP();
   uint16_t remotePort();
 
-  void setTimeout(unsigned long timeout);
+  void setSocketTimeout(unsigned long timeout);
 
   friend class MbedServer;
   friend class MbedSSLClient;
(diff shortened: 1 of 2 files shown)


JAndrassy:
it is not a connection timeout as setConnectionTimeout should do!
it is timeout on other operations, not on connect

libraries/SocketWrapper/src/MbedClient.h
```diff
@@ -97,7 +97,7 @@ class MbedClient : public arduino::Client {
   IPAddress remoteIP();
   uint16_t remotePort();
 
-  void setTimeout(unsigned long timeout);
+  void setConnectionTimeout(unsigned long timeout);
```
facchinm, line 100: Let's rename as `setSocketTimeout` as it's indeed more meaningful

facchinm (changes requested):

Issue #130, opened 2021-02-07
PluggableUSBModule::getProductVersion() not defined


tttapa:
Commit 301d69d045d9f3e636ec4f8a1f8d5b02b2cb0e3f introduces a `PluggableUSBModule::getProductVersion()` method declaration, but it is not defined in the base class, resulting in a linker error when derived classes don't override it.

I believe this method should be pure virtual:
```cpp
virtual uint8_t getProductVersion() = 0;
```
Is there a guideline for which value third-party libraries should use? Should they just set the next available bit (16)? Is there an upper bound?

From the name of the function alone, it was not really clear to me how the function is intended to be used. The links in the commit message explain the rationale behind it, but it might be useful to have a comment or docstring in the PluggableUSBModule class documenting how and why to use it when defining a module.

facchinm:
@tttapa thanks for pointing that out! I didn't think there were any other module in development when I pushed it so I took the freedom to make a breaking change, my bad.
Feel free to post a PR for the pure virtual method so we keep the attribution; about the value to be assigned, it would be cool if it could be automaticlly generated in some way, but in the meantime using the next bit (16) is perfectly fine.

tttapa:
@facchinm no problem, it's easy enough to fix :)

I've opened a pull request with the proposed change. When the method is not implemented in a derived PluggableUSBModule, you now get a clear compiler message instead of a linker error:
```
In file included from /home/pieter/Arduino/libraries/Control-Surface/src/MIDI_Interfaces/USBMIDI/USBMIDI.hpp:55:0,
                 from /home/pieter/Arduino/libraries/Control-Surface/examples/3. MIDI Interfaces/MIDI_Pipes-Routing/MIDI_Pipes-Routing.ino:14:
/home/pieter/Arduino/libraries/Control-Surface/src/MIDI_Interfaces/USBMIDI/USBMIDI_Arduino_mbed.hpp:19:22: error: cannot declare field 'CS::USBDeviceMIDIBackend::backend' to be of abstract type 'CS::PluggableUSBMIDI'
     PluggableUSBMIDI backend;
                      ^~~~~~~
In file included from /home/pieter/Arduino/libraries/Control-Surface/src/MIDI_Interfaces/USBMIDI/USBMIDI_Arduino_mbed.hpp:4:0,
[...]

PR #771, opened 2023-11-23
USB: add pure specifiers and emit vtable

ardnew:
Issue #130 correctly identifies a newly-added method as pure virtual and fixes the declaration. However, for some reason it didn't address all of the other virtual methods in that same class (`PluggableUSBModule`) that do not define a default implementation.

The only virtual method that has a default implementation is provided inline in the class interface:

```c++
    virtual void callback_reset() {};
```

These issues combined prevent the compiler from being able to emit a vtable for the `PluggableUSBModule` class, thus preventing users from correctly subclassing it or any one of its derived classes such as `USBCDC`, `USBHID`, `USBMIDI`, etc. Refer to the following answer on StackOverflow for a detailed explanation of the issue:

https://stackoverflow.com/a/57504289/1054397

This PR adds the pure specifier (`= 0`) to all of the virtual methods in this class that do not have a default implementation. It also moves the default empty definition of `virtual void callback_reset()` to the class definition in `USB/PluggableUSBDevice.cpp` so that this class complies completely with the criteria for emitting a vtable.

> #### Note
>
> The error that I was encountering prior to these changes was pretty
> cryptic (from PlatformIO):
>
> ```txt
> .pio/build/hid/src/target.cpp.o: In function `foo()':
> USBHID/src/PluggableUSBHID.h:53: undefined reference to
>     `vtable for arduino::internal::PluggableUSBModule'
> .pio/build/hid/src/target.cpp.o: In function `foo()':
> foo.hpp:100: undefined reference to
>     `vtable for arduino::internal::PluggableUSBModule'
> collect2: error: ld returned 1 exit status
> *** [.pio/build/hid/firmware.elf] Error 1
> ```
>
> Even stranger, the error would only be generated with a debug build;
> i.e., the only difference in command-line arguments was the additional
> CFLAGS of `-Og -g2 -ggdb2`. Without the debug flags, my project was
> building without error.
>
> With the changes in this PR, my project now builds with and without
> these additional debug flags. Further verification was performed by
> testing the example sketches `Keyboard`, `KeyboardRaw`, and `Mouse`
> from library `USBHID` as well as using the core `Serial` object for
> ordinary USB serial I/O (`USBCDC`).

### DIFF
--- a/cores/arduino/USB/PluggableUSBDevice.cpp
+++ b/cores/arduino/USB/PluggableUSBDevice.cpp
@@ -60,6 +60,10 @@ uint32_t arduino::internal::PluggableUSBModule::write_finish(usb_ep_t endpoint) 
     return PluggableUSBD().write_finish(endpoint);
 }
 
+void arduino::internal::PluggableUSBModule::callback_reset()
+{
+}
+
 arduino::PluggableUSBDevice::PluggableUSBDevice(uint16_t vendor_id, uint16_t product_id)
     : USBDevice(get_usb_phy(), vendor_id, product_id, 1 << 8)
 {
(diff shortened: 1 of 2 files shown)


facchinm:
LGTM! Thanks!

PR #760, opened 2023-11-15
Update Wire.cpp

watershade:
flush function have been never be implemented. But it is very useful. So I addvice to implement it. Thanks.

PR #773, opened 2023-11-24
Implement the Wire.flush() funtion

watershade:
1. implement flush of wire. (flush functions have never been implemented before. But it is useful.)

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -136,6 +136,7 @@ int arduino::MbedI2C::peek() {
 }
 
 void arduino::MbedI2C::flush() {
+	rxBuffer.clear();
 }
 
 #ifdef DEVICE_I2CSLAVE


watershade:
Wire.flush() is useful, if I haven't read  obsolete data from it. I will get a wrong number via Wire.available() because the size have include the obsolete data. And If I use Wire.read(), the data I read out is include the obsolete data also. Yes, I can read the obsolete data out via Wire.read(), but it is not necessary. If Wire.flush() is I implemented, It will become perfect.
Thanks！

watershade:
@facchinm Hi Martino, I don't know if you time to check this merge. Thanks.

tokiedokie:
Duplicate of https://github.com/arduino/ArduinoCore-mbed/pull/760

pennam:
Flush should be intended as a way to send all data out of the TX buffer not to cleanup the RX. See Serial.flush() https://www.arduino.cc/reference/it/language/functions/communication/serial/flush/

tokiedokie:
Should the RX buffer cleanup API be called `clear()`?

pennam:
> Should the RX buffer cleanup API be called `clear()`?

I don't think there is a standard Arduino API, but personally i think is a more appropriate name.

watershade:
@pennam Thanks for your reply. I will think about it and find a better way. Maybe next time I will add unlock function together.

Issue #772, opened 2023-11-23
New release breaks offloading of camera read to M4 core


Bexin3:
Hello, I have managed to get a code with m4 core reading the camera instead of m7, as unfortunately, direct memory access isn't used, and thus, lot of processing time is taken by reading the camera. However ever since the new release, the code no longer works, which may be a result of some of the M4 changes and is certainly unfortunate.
The link to the code;
https://forum.arduino.cc/t/gigacameratodisplay-offloading-camera-reading-to-the-m4-core/1191858/3

PR #741, opened 2023-10-23
Giga: compatibility fixes with M4 builds

facchinm:
Fixes https://github.com/arduino/ArduinoCore-mbed/issues/726 and supersedes https://github.com/arduino/ArduinoCore-mbed/pull/740

PR #774, opened 2023-11-24
STM32H747: SDRAM on M4 fixes

facchinm:
Should fix https://github.com/arduino/ArduinoCore-mbed/issues/772 (see https://github.com/arduino/ArduinoCore-mbed/pull/741#issuecomment-1825450023 ).

Coexistence between M4 and M7 got broken by #741 , alongside `printf` functionality in M4 (sicne default console is not defined anymore)

@Bexin3  can you test with your project? I tried it and it seems to work but the tearing is "important" (and it's not a regression as far as I can tell)


Bexin3:
I will test this in a few hours when I am back home.
Regarding the tearing, that one also happened in older versions, but it wasn't fixed which I thing some of the update notes mentioned it should be fixed. The tearing is a vertical stripe, and its not a vsync issue but particles appear on both sides of the tear, similar behaviour sometimes happened when I had an incorrect if function writing into a buffer, for some reason pixels would sometimes jump over the place. I think that the tear leaves some pixels unapdates on both sides while some get updated but honestly I can't say for certain, I can take a photo when home.

Bexin3:
Is there anywhere I could get this compiled? I am afraid I am not really good at understanding compilation stuff

PR #783, opened 2023-12-01
Ethernet.h remove unimplemented methods

JAndrassy:
* socketAddressFromIpAddress is  the base class 
* setHostname causes linker error. implementation is not possible

### DIFF
--- a/libraries/Ethernet/src/Ethernet.h
+++ b/libraries/Ethernet/src/Ethernet.h
@@ -91,8 +91,6 @@ public:
 
   void MACAddress(uint8_t *mac_address);
 
-  void setHostname(const char *name);
-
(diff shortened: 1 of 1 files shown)


facchinm:
Thanks for the contribution, they were stray functions from the very old initial implementation. Merging!

JAndrassy:
@facchinm Martino, to be honest, the networking libraries of this core look unfinished. There are too many loose ends. I will add more PR, but without [working client copying](https://github.com/arduino/ArduinoCore-mbed/pull/768) it is not even possible to test MbedServer.
And I still don't have hardware (or software in form of an EMAC driver for a cheap WiFi) to test the WiFi library, but in code I see that there will be problems with static IP configuration

Issue #758, opened 2023-11-12
GIGA R1 - analogRead(A12) returns A8 value


megacct:
and analogRead(A13) returns A9 value. However, using Adafruit advancedAnalog library the correct values for A12 and A13 are returned

facchinm:
Mmmmh no, I retested just to make sure and it works as intended (reads A13/DAC1 value when `analogRead(A13)` is called).
Are you sure you are not using `digitalRead()` ? Because pins from A8 to A11 are analog only and get skipped in the count when using as digital https://github.com/arduino/ArduinoCore-mbed/blob/main/variants/GIGA/pins_arduino.h#L54-L55 since they can't be used as digital pins.

facchinm:
Got it, I've been able to replicate.
@Rocketct can you take a look?

Rocketct:
sure @facchinm

Rocketct:
https://github.com/arduino/ArduinoCore-mbed/pull/784 this pr should fix the issue

white-ar:
Hello, I've just come across the issue with a Giga R1.  Upgrading to V 4.1.5 of the board library _temporarily_ fixed the issue, however it has returned.  Here is the very simple sketch:

@facchinm @Rocketct is there a way of addressing A9 using MBED calls?  (as a work around).  Thank you. 

`
#include "Arduino.h"
#include "RPC.h"

#define SERVO_POT_1     A9

void setup() {
  // put your setup code here, to run once:
  RPC.begin();

  analogReadResolution(16); //GIGA has 16 bit ADC  (0-65535)  

}

void loop() {
  // put your main code here, to run repeatedly:

int temp = analogRead(SERVO_POT_1);



}`

Error message:
C:\...\AppData\Local\Temp\arduino\sketches\D2129B132D44D2F5F57EC30CA80C2E2B\sketch\sketch_jul25a.ino.cpp.o: In function `loop':
C:\...\AppData\Local\Temp\.arduinoIDE-unsaved2024625-43476-ja9cee.pq8d\sketch_jul25a/sketch_jul25a.ino:20: undefined reference to `A9'
C:\...\AppData\Local\Temp\.arduinoIDE-unsaved2024625-43476-ja9cee.pq8d\sketch_jul25a/sketch_jul25a.ino:20: undefined reference to `analogRead(PureAnalogPin)'
collect2.exe: error: ld returned 1 exit status

[...]

PR #784, opened 2023-12-01
fix ADC management for pure analog on GIGA

Rocketct:
This PR allows to fix the management of the pure analogs on GIGA, the issue was that the pure analog pins have not been considered on the **g_APinDescription** array, this results on miss an offset of 4 elements when a **g_AAnalogPinDescription's pin** was select in the **analogRead()** in **wiring_Analog.cpp**.
The consequence is that the macros involved in Pin and ADC selection, could potentially identify an ADC assigned to the pure analog pins as the ADC channel for all those analog listed after the pin A11 in **g_AAnalogPinDescription array** in the **variant.cpp** and than read the ADC channel of the pure analog pins instead of the correct one (e. g. A8 for A12, A9 for A13 and so on and so forth).

The Fix split the two kind of analog pin in order that the correct offset is provided during reading time and manage the pure analog and the other analogs pin as a different entities inside the pins management.

This PR fixes https://github.com/arduino/ArduinoCore-mbed/issues/758

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -84,6 +84,7 @@ typedef struct _AnalogPinDescription AnalogPinDescription;
 extern PinDescription g_APinDescription[];
 extern AnalogPinDescription g_AAnalogPinDescription[];
 extern AnalogPinDescription g_AAnalogOutPinDescription[];
+extern AnalogPinDescription g_pureAAnalogPinDescription[];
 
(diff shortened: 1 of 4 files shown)


njadNissi:
I have the most updated version of this library in my ArduinoIDE, I checked that all the code is the same. however calling `anologRead(A8)` or of any of the pureAnalogPins still doesn't work. could you provide an example sketch to show how it should be used?

PR #785, opened 2023-12-01
Socket download function improvements

andreagilardoni:
In this PR I split the download function into 2 versions:
- version 1: the old one that downloads a file and stores it on the fs
- version 2: download a file and allow the user to specify a callback to handle the incoming body buffer
This is required to perform LZSS decompression on the fly


facchinm:
LGTM! Thanks!

PR #791, opened 2023-12-05
Ethernet - allow return to DHCP after `begin` with static IP

JAndrassy:
the `begin` for DHCP didn't set dhcp to true. the same function was used internally from `begin` for static IP. 
this PR separates a private `_begin` from `begin` for DHCP and sets dhcp to true in the public `begin` for DHCP.

tested with [LegacyEthernetTest](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/LegacyEthernetTest/LegacyEthernetTest.ino)


JAndrassy:
@facchinm I added a description yesterday. this is a simple PR

JAndrassy:
this would not work with the Arduino_ConnectionHandler library

PR #793, opened 2023-12-06
SocketWrapper - MbedServer modernization (without available() and Print)

JAndrassy:
MbedServer didn't manage clients for proper available() and print-to-all-clients. Now available() in derived server classes WiFiClient and EthernetClient is deprecated and accept() is added with the same implementation. Inheriting from Print (Server) is removed. `write` methods for Print implementation are removed. They never worked.

New are constructor without parameters, begin with parameter port and end() as in the new WiFiS3 library.

The ChatServer examples never worked because they relay on Processing style server.available() and print-to-all-clients.. Removed. 

ESP32 core and RP2040 core libraries too don't implement Processing style Servers. Processing style WiFiServer and EthernetServer for these cores and Mbed core is implemented in my NetAPIHelpers library. In my networking libraries WiFiEspAT and EthenetENC the print-to-all-clients is in separate class and in next major version I plan to remove server.available() and point to NetAPIHelpers Server.

This replaces PR https://github.com/arduino/ArduinoCore-mbed/pull/750 and https://github.com/arduino/ArduinoCore-mbed/pull/751

overview of Server implementations in libraries https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#server-class


JAndrassy:
@facchinm  @pennam

now this is my last essential PR for Mbed Core networking. It does changes proposed by Alessandro Ranellucci, but including the consequences

https://github.com/arduino/ArduinoCore-mbed/pull/750  adds accept() in WiFiServer

consequences:
* add accept in EthernetServer
* deprecate `available`
* modify examples to use accept() and remove Ethernet examples which require true server.available() (WiFi lib doesn't have them)
* remove unimplemented Print inheritance and write functions, because print-to-all-clients is tied with `server.available()`

https://github.com/arduino/ArduinoCore-mbed/pull/751 adds server.end() in WiFiServer

consequences:
* add end() in EthernetServer

other related server classes modernization as in WiFiS3:
* add `begin` with parameter `port`` and constructor without parameters
* add operator bool

pennam (commented):
@JAndrassy is this still true? the whole system works but i cannot see the clients input in the serial monitor.

https://github.com/JAndrassy/ArduinoCore-mbed/blob/18b696b224f8c621cd95fb099e63e25c68245c96/libraries/Ethernet/examples/AdvancedChatServer/AdvancedChatServer.ino#L7

JAndrassy:
it is just a comment which is not true. it is same in the Ethernet library from where the example originates

PR #801, opened 2023-12-18
variants/GIGA: Fix RPMSG endpoint buffer size.

iabdalkader:
This needs to match the buffer size used in variants/GENERIC_STM32H747_M4.


iabdalkader:
This was combined in another PR.

PR #809, opened 2023-12-22
Fix DNS get methods

manchoz:
This PR fixes the `::dnsServerIP()` and  `::dnsIP()` methods to retrieve the DNS server configuration properly.

The issue is due to [NetworkInterface.cpp#L111-L112](https://github.com/manchoz/mbed-os/blob/f6d6cdfcefc77d0bf333cf60812a1d0b7282a5f5/connectivity/netsocket/source/NetworkInterface.cpp#L111-L112) ultimately requiring the interface name because of [lwip_dns.c#L505-L506](https://github.com/manchoz/mbed-os/blob/f6d6cdfcefc77d0bf333cf60812a1d0b7282a5f5/connectivity/lwipstack/lwip/src/core/lwip_dns.c#L505-L506)

### DIFF
--- a/libraries/SocketWrapper/src/SocketHelpers.cpp
+++ b/libraries/SocketWrapper/src/SocketHelpers.cpp
@@ -50,14 +50,18 @@ arduino::IPAddress arduino::MbedSocketClass::gatewayIP() {
 arduino::IPAddress arduino::MbedSocketClass::dnsServerIP() {
   SocketAddress ip;
   NetworkInterface* interface = getNetwork();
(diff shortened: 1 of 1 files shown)


JAndrassy:
for curios this is the implementation of get_dns_server in NetworkInterface.cpp in Mbed OS
```
nsapi_error_t NetworkInterface::get_dns_server(int index, SocketAddress *address, const char *interface_name)
{
    return get_stack()->get_dns_server(index, address, interface_name);
}
```

PR #814, opened 2023-12-28
Refactor and fix RPC library.

### LINKED ISSUE #802: RPC library deadlock and/or vring memory corruption on Giga R1 (possibly other boards too).
The attached sketch generates a random block of data on each core, and sends it to the other one to hash it. It then reads back the hash and compares it to the one generated locally. 
If you run the sketch you'll see that it deadlocks and/or corrupts the memory in less than a minute or so. The Red LED blink means one side received a bad md5 checksum. 

NOTE: you must fix the endpoint buffer size first before testing this, as in PR https://github.com/arduino/ArduinoCore-mbed/pull/801

[rpc_md5sum.zip](https://github.com/arduino/ArduinoCore-mbed/files/13704952/rpc_md5sum.zip)

iabdalkader:
Main changes in this PR:

- Fix the memory corruption caused by the wrong MPU region configuration, which enabled caching of the shared memory region, without any cache maintenance in libmetal.
- Fix the MPU region's TEX level, which was setting the shared memory region as a device and/or strongly-ordered memory, which caused usage faults on unaligned accesses.
- Fix a deadlock that occurs if the main thread (sketch) makes a remote call and holds the virtqueue's mutex, and then gets preempted by the event thread, which in turn tries to acquire the same mutex. Note that libmetal's mutex implementation doesn't yield the CPU, it can be changed to do so, but for now the queues are protected with an OS mutex.
- Fix RPCLIB msgpack unpacker memory overflow if RPMSG endpoint buffer > 512.
- Optimize RPCLIB msgpack unpacker usage by allocating once and extending on demand.
- Remove extra request/response threads, locking and callbacks, that were causing all sorts of concurrency issues, and were detrimental to performance.
- Remove the redundant copy of messages to ring buffer, to boost the performance.
- Remove many smaller dynamic allocation to improve the performance.
- Use a single endpoint for message packs requests/responses ("RPC"), and another one dedicated for raw (non-packed) data.
- General refactoring and cleanups, removal of dead code and variables, debugging code, etc...
- Add MD5 checksum example and RAW ping-pong example.
- Fixes #802.

This PR depends on the following:

- [ ] #799 
- [ ] #803

### DIFF
--- a/libraries/RPC/examples/Basic_AddSub/basic_add_sub.ino
+++ b/libraries/RPC/examples/Basic_AddSub/basic_add_sub.ino
@@ -1,0 +1,25 @@
+#include "RPC.h"
+
+int add(int a, int b) {
+  return a + b;
+}
+
+void setup() {  
+  RPC.begin();
+  RPC.bind("add", add);
+  pinMode(LEDG, OUTPUT);
+}
+
+void loop() {
+  static size_t loop_count = 0;
+  // Blink every 512 iterations
+  if ((loop_count++ % 512) == 0) {
+    digitalWrite(LEDG, LOW);
+    delay(10);
+    digitalWrite(LEDG, HIGH);
+    delay(10);
+  }
+  int res = RPC.call("add", 1, 2).as<int>();
+  RPC.call("sub", res, 1).as<int>();
+  delay(250);
+}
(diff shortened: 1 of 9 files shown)


facchinm:
Manually merged in `main`

PR #816, opened 2024-01-02
WiFi - make standard `config` functions work

JAndrassy:
- make standard Arduinp `config` methods work for WiFi
- fill automatic dns, gw and netmask values
- allow reconnecting with static IP
- allow reconnecting with DHCP after static IP

PR https://github.com/arduino/ArduinoCore-mbed/pull/809 is related to return the DNS server IP

[overview](https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#wifi-station-network-interface) of WiFi API in libraries


facchinm:
Wonderful, thank you so much for every PR :hugs:

PR #832, opened 2024-01-25
Support for Pico second serial port

o-gent:
Hi!

To my dismay I found Serial2 didn't work when I started messing with the Pico. 

This change adds the pin definition required for Serial 2 to GPIOs 8 & 9. I used these since the default definition for i2c is on 4&5.

I've tested this on my setup and successfully received data on Serial2.


o-gent:
Hi @facchinm, I see you've merged a bunch of the previous pull requests - do you see any issue with this proposal?

PR #841, opened 2024-02-14
fix(Ethernet): missing return statement in begin with mac, timeout and responseTimeout

trik:


JAndrassy:
error from my PR https://github.com/arduino/ArduinoCore-mbed/pull/823. 
sorry

PR #849, opened 2024-03-05
SocketWrapper - use DNS of specific netif, not global

JAndrassy:
DNS IP address assigned by DHCP was not used.
Instead DNS servers hardcoded in Mbed were used.

error report https://forum.arduino.cc/t/ethernet-dhcp-issue/1231893

### DIFF
--- a/libraries/SocketWrapper/src/MbedClient.cpp
+++ b/libraries/SocketWrapper/src/MbedClient.cpp
@@ -129,7 +129,7 @@ int arduino::MbedClient::connect(IPAddress ip, uint16_t port) {
 int arduino::MbedClient::connect(const char *host, uint16_t port) {
   SocketAddress socketAddress = SocketAddress();
   socketAddress.set_port(port);
-  getNetwork()->gethostbyname(host, &socketAddress);
+  SocketHelpers::gethostbyname(getNetwork(), host, &socketAddress);
   return connect(socketAddress);
 }
 
@@ -199,7 +199,7 @@ int arduino::MbedClient::connectSSL(const char *host, uint16_t port, bool disabl
 
   SocketAddress socketAddress = SocketAddress();
   socketAddress.set_port(port);
-  getNetwork()->gethostbyname(host, &socketAddress);
+  SocketHelpers::gethostbyname(getNetwork(), host, &socketAddress);
   return connectSSL(socketAddress);
 }
 
--- a/libraries/SocketWrapper/src/MbedUdp.cpp
+++ b/libraries/SocketWrapper/src/MbedUdp.cpp
@@ -63,7 +63,7 @@ int arduino::MbedUDP::beginPacket(IPAddress ip, uint16_t port) {
 int arduino::MbedUDP::beginPacket(const char *host, uint16_t port) {
   _host = SocketAddress(host, port);
   txBuffer.clear();
-  getNetwork()->gethostbyname(host, &_host);
+  SocketHelpers::gethostbyname(getNetwork(), host, &_host);
   //If IP is null and port is 0 the initialization failed
   return (_host.get_ip_address() == nullptr && _host.get_port() == 0) ? 0 : 1;
 }
--- a/libraries/SocketWrapper/src/SocketHelpers.cpp
+++ b/libraries/SocketWrapper/src/SocketHelpers.cpp
@@ -17,7 +17,7 @@ String arduino::MbedSocketClass::macAddress() {
 
 int arduino::MbedSocketClass::hostByName(const char* aHostname, IPAddress& aResult) {
   SocketAddress socketAddress = SocketAddress();
-  nsapi_error_t returnCode = getNetwork()->gethostbyname(aHostname, &socketAddress);
+  nsapi_error_t returnCode = gethostbyname(getNetwork(), aHostname, &socketAddress);
   nsapi_addr_t address = socketAddress.get_addr();
   aResult[0] = address.bytes[0];
   aResult[1] = address.bytes[1];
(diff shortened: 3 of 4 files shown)


pennam:
@manchoz

manchoz:
Uh, thanks. I will check it ASAP.

JAndrassy:
it is the same thing as in PR https://github.com/arduino/ArduinoCore-mbed/pull/809 but here to really use the DNS servers we set and get

manchoz:
@JAndrassy Did you try it?

Currently, you need to use the following code to apply the DNS received from the DHCP server to the network stack.

```
void onNetworkConnect()
{
  Serial.println(">>>> CONNECTED to network");
  Serial.println("Setting Ntp Time");
  setNtpTime(); // Everytime the network is connected, we sync the system time
  Serial.print("Assigned IP:");
  Serial.println(Ethernet.localIP());

/* It requires core patch from #809 
  auto dnsAddress = Ethernet.dnsServerIP();
  Serial.println(dnsAddress);

  auto primaryDnsAddress = Ethernet.dnsIP(0);
  Serial.println(primaryDnsAddress);
  auto secondaryDnsAddress = Ethernet.dnsIP(1);
  Serial.println(secondaryDnsAddress);
*/

  auto net = Ethernet.getNetwork();

  char name[5] {};
  net->get_interface_name(name);
  Serial.print("IF name: ");
  Serial.println(name);

  SocketAddress primaryDNS;
  SocketAddress secondaryDNS;

  net->get_dns_server(0, &primaryDNS, name);
  if (primaryDNS) {
    Serial.print("Primary DNS Server: ");
    Serial.println(ipAddressFromSocketAddress(primaryDNS));
    Serial.println("Setting Primary DNS Server...");
    net->add_dns_server(primaryDNS, nullptr);
  }

  net->get_dns_server(1, &secondaryDNS, name);
  if (secondaryDNS) {
    Serial.print("Secondary DNS Server: ");
    Serial.println(ipAddressFromSocketAddress(secondaryDNS));
    Serial.println("Setting Secondary DNS Server...");
    net->add_dns_server(secondaryDNS, nullptr);
  }
}
  

IPAddress ipAddressFromSocketAddress(SocketAddress socketAddress) {
  nsapi_addr_t address = socketAddress.get_addr();
  return IPAddress(address.bytes[0], address.bytes[1], address.bytes[2], address.bytes[3]);
}
```

The patch that we would need should be able to do this stuff when we receive the DNS configuration from the DHCP server, otherwise we will end using https://github.com/manchoz/mbed-os/blob/f6d6cdfcefc77d0bf333cf60812a1d0b7282a5f5/connectivity/netsocket/source/nsapi_dns.cpp#L116.

JAndrassy:
I tried it with WiFi. I didn't try Ethernet. 
in WiFi.begin `wifi_if->set_dhcp` the DNS servers are set to the specific interface. it should work the same in Ethernet.begin

papakpmartin:
Is this still broken in Release 4.1.1? I'm having a heckuva time on an Opta with very simple stuff that Just Works™ on ESP32. I'm attaching a simple case of what I think should be working fine but isn't. This code does _print_ the correct DNS server address (so #809 is fixed) but it can't find the MQTT broker (which can by found by other devices on the same network by the same hostname using identical code).

[broken-dns-example.txt](https://github.com/arduino/ArduinoCore-mbed/files/14830453/broken-dns-example.txt)

markwestm:
@papakpmartin and I are working on the same project. I am also having DNS issues running 4.1.1 on the Opta. Specifically, I cannot get a local DNS via DHCP. The IP's I get are shown below. My code file is attached below. I even updated the four files associated with this PR manually in the packages folder. Same issue. I also tried the onNetworkConnect() function recommended by @manchoz as well as the work-around in https://forum.arduino.cc/t/ethernet-dhcp-issue/1231893.

ETH: Local IP: 192.168.0.179
ETH: DNS Server: 68.105.28.11

[dns-code.txt](https://github.com/arduino/ArduinoCore-mbed/files/14842952/dns-code.txt)

markwestm:
This is an update to my comment from two days ago. I discovered that setting the DNS server IP is not possible via a DHCP ethernet connection. I found a work-around to my problem by doing the following.

1. Get and store the device IP address and gateway IP address from the original DHCP ethernet connection.
2. Terminate the DHCP ethernet connection.
3. Establish a new non-DHCP ethernet connection by passing the MAC address, device IP, and gateway IP to the ethernet begin function.
4. Set the DNS server IP address to the gateway IP address retrieved in step 1.

Regards, Mark

JAndrassy:
@markwestm pleas try to to replace in Ethernet.cpp
```
  eth_if->add_dns_server(_dnsServer1, nullptr);
```
to
```
  char if_name[5];
  eth_if->get_interface_name(if_name);
  eth_if->add_dns_server(_dnsServer1, if_name);
```

JAndrassy:
@manchoz which way?
1) write and read DNS server IP with netif name parameter everywhere?
2) copy DHCP set DNS IP to global DNS server IPs?

your PR https://github.com/arduino/ArduinoCore-mbed/pull/809 would indicate 1) and the Arduino API DNS functions are methods of the specific network interface so it is simple to manage them that way.

with 2) there are a few problems: 
* there are limited number of DNS servers in the global registry (2?)
* the DNS server IP management from the sketch with Arduino function would not be possible, because WiFi and Ethernet would overwrite each others records if used at the same time

JAndrassy:
@markwestm I think the changes from this PR solve your problem but for some reason they where not applied to your test

markwestm:
@JAndrassy Is there any information (code snippets, serial monitor output, etc.) I can provide to help determine if these changes were applied to my test. Thanks for your support.

Regards, Mark

JAndrassy:
> s there any information (code snippets, serial monitor output, etc.) I can provide to help determine if these changes were applied to my test. 

without these changes you can see in Wireshark that it uses 8.8.8.8 for a DNS request. With the changes it uses the DHCP assigned DNS server IP

markwestm:
@JAndrassy Would you mind posting those changes again? I don't see them in the thread. Thanks in advance.

papakpmartin:
@JAndrassy I can confirm that the changes in this PR are working. Thanks so much!

@pennam and @manchoz, it would be awesome if this could get merged in... thanks! :)

PR #856, opened 2024-03-26
Added Portenta C33 (0068) to post_install.sh

hybotics:
I added the Portenta C33 (0068) to postinstall.sh so the udev rules would include it. This fixes the upload error 74 in Linux.


JAndrassy:
Portenta C33 is not in this boards package

facchinm:
C33 handled here https://github.com/arduino/ArduinoCore-renesas/blob/main/post_install.sh

PR #859, opened 2024-04-04
NICLA VISION TLS skip server certificate date check

pennam:
This check was enabled only for NICLA VISION, and connect was failing because system time is not automatically configured. On other variants date checks are not enabled because `MBEDTLS_HAVE_TIME_DATE` is not defined.


pennam:
WiFiSSLClient :green_heart: 
ArduinoCloud :green_heart:

PR #869, opened 2024-04-17
GSM: remove possibility to set timeout from constructor

pennam:
`GSMClient` and `GSMSSLClient` where the only two Clients that have the possibility to set timeout from constructor, this was handy but not required because user can use `setSocketTimeout()` method to configure timeout.

Furthermore after merging #768 the GSMClient has been broken because we are calling a virtual function `getNetwork()` from the constructor that is calling `__cxa_pure_virtual()` and then `abort()` resulting in a red blink of death.


pennam:
Portenta H7 + THALES - GSMClient :+1: 
Portenta H7 + THALES - GSMSSLClient :+1: 
Portenta H7 + THALES - ArduinoIoTCloud :+1: 

EdgeControl + THALES - GSMClient :+1: 
EdgeControl + THALES - GSMClient :+1: 
EdgeControl + THALES - GSMClient :+1:

PR #874, opened 2024-04-22
WiFi: Support for hidden networks and misc fixes.

### LINKED ISSUE #819: .begin() for WEP is in Wifi.h but is not implemented in Wifi.cpp.
I have been asked to raise an issue here but I have never done this before so forgive any errors in formatting, etc.

I have a Giga R1 and the accompanying display and wanted to get the time from an NTP server.  I therefore followed the official Arduino examples shown here: [https://docs.arduino.cc/tutorials/giga-r1-wifi/giga-wifi](url)

I found that I could make a WPA connection using...

`status = WiFi.begin(ssid, pass);`

...but that when I tried to make a WEP connection by adding the network index number like this:

`status = WiFi.begin(ssid, keyIndex, pass);`

...I got a compilation error.

I raised this on the Arduino Forum here  [https://forum.arduino.cc/t/giga-wifi-not-working/1207373](url) and a helpful person called stevebeswick replied saying that "_the .begin() for WEP is in Wifi.h but is not implemented in Wifi.cpp_" and suggested that I should raise an issue here about it - which I have now done.

I hope this provides all necessary information.

### LINKED ISSUE #855: Connectivity to hidden WiFi networks
Hi,

I attempted to connect my Arduino Giga R1 WiFi board to my secure central WiFi, but they are not linking up. Later I found out that only Arduino Giga R1 WiFi doesn't connect to WiFi, but the simple ESP8266 modules I connected to my Mega board get WiFi from the hidden SSID.

I would like to request the team to update the GIGA R1 WiFi board library, allowing it to connect to Wi-Fi access points that do not broadcast an SSID. I can't modify AP and without the Giga being able to connect to hidden WiFi networks, I can't use my two Gigas for my important project.

Thank you in advance for considering my request.

iabdalkader:
- Add support for connecting to hidden networks.
- Implement `begin()` for OPEN security.
- Deprecate WEP security.
- Add sanity checks to all functions accepting network index from the user.
- Fixes #855
- Fixes #819

Note that for hidden networks, the BSSID is Not available, since it was not found in a scan. However, it can still be retrieved with:
```C
static_cast<WhdSTAInterface *>(wifi_if)->get_bssid(bssid);
```
I'm not sure why we're not using that, maybe for portability ?

Also not that `uint8_t* BSSID(uint8_t* bssid);` returns a reversed MAC (due to a legacy bug in Nina)  however `uint8_t* BSSID(uint8_t networkItem, uint8_t* bssid);`  does Not return a reversed MAC. I didn't change that.

### DIFF
--- a/libraries/WiFi/src/WiFi.cpp
+++ b/libraries/WiFi/src/WiFi.cpp
@@ -3,6 +3,39 @@
 #define SSID_MAX_LENGTH 32
 #define SSID_MAX_COUNT  64
 
+static uint8_t sec2enum(nsapi_security_t sec) {
+  switch (sec) {
+    case NSAPI_SECURITY_NONE:
+      return ENC_TYPE_NONE;
+    case NSAPI_SECURITY_WEP:
+      return ENC_TYPE_WEP;
+    case NSAPI_SECURITY_WPA:
+      return ENC_TYPE_TKIP;
+    case NSAPI_SECURITY_WPA2:
+      return ENC_TYPE_CCMP;
+    case NSAPI_SECURITY_WPA_WPA2:
+      return ENC_TYPE_CCMP;
+    case NSAPI_SECURITY_UNKNOWN:
+    default:
+      return ENC_TYPE_AUTO;
+  }
+}
+
+static nsapi_security_t enum2sec(wl_enc_type sec) {
+  switch (sec) {
+    case ENC_TYPE_NONE:
+      return NSAPI_SECURITY_NONE;
+    case ENC_TYPE_WEP:
+      return NSAPI_SECURITY_WEP;
+    case ENC_TYPE_TKIP:
+      return NSAPI_SECURITY_WPA;
+    case ENC_TYPE_CCMP:
+      return NSAPI_SECURITY_WPA_WPA2;
+    default:
+      return NSAPI_SECURITY_UNKNOWN;
+  }
+}
+
 bool arduino::WiFiClass::isVisible(const char* ssid) {
   for (int i = 0; i < SSID_MAX_COUNT; i++) {
     if (strncmp(ap_list[i].get_ssid(), ssid, SSID_MAX_LENGTH) == 0) {
@@ -13,7 +46,7 @@ bool arduino::WiFiClass::isVisible(const char* ssid) {
   return false;
 }
 
-int arduino::WiFiClass::begin(const char* ssid, const char* passphrase) {
+int arduino::WiFiClass::begin(const char* ssid, const char* passphrase, wl_enc_type security) {
   if (wifi_if == nullptr) {
     return 0;
   }
@@ -21,10 +54,13 @@ int arduino::WiFiClass::begin(const char* ssid, const char* passphrase) {
   wifi_if->attach(&arduino::WiFiClass::statusCallback);
 
   scanNetworks();
-  // use scan result to populate security field
-  if (!isVisible(ssid)) {
-    _currentNetworkStatus = WL_CONNECT_FAILED;
-    return _currentNetworkStatus;
+
+  if (isVisible(ssid)) {
+    // Set the network security mode from the scan result.
+    _security = ap_list[connected_ap].get_security();
+  } else {
+    // For hidden networks, the security mode must be set explicitly.
+    _security = enum2sec(security);
   }
 
   wifi_if->set_dhcp(!_useStaticIP);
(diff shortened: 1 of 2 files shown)


sebromero (changes requested):
Good stuff, thanks Ibrahim. I suggest for the sake of portability und user friendliness to not expose non-arduino types to the user. Can we define our own enum for the security type?

iabdalkader:
> I suggest for the sake of portability und user friendliness to not expose non-arduino types to the user. Can we define our own enum for the security type?

We seem to have our own enum and I've reused it, that's why I added `enum2sec`. The enum returned/expected is `ENC_TYPE_xxx`.

sebromero:
> > I suggest for the sake of portability und user friendliness to not expose non-arduino types to the user. Can we define our own enum for the security type?
> 
> We seem to have our own enum `wl_enc_type` and I've reused it, that's why I added `enum2sec`. The enum returned/expected is `ENC_TYPE_xxx`.

Hmm, okay. The enum doesn't look very Arduino-ish, but then that's a different problem 😅

JAndrassy:
@iabdalkader I did this a few weeks ago, but then I didn't do a PR because:
* getEncryptionType returned AUTO if encryption type was specified at connect and not scanned
* If I remember it right, AUTO encryption type at connect worked, so the encryption type parameter in `begin` was maybe not necessary
* for WEP different connect parameters are required (key index?)
* maybe the results I saw where caused by the fact that I don't have Portenta or Giga. To test Mbed WiFi I use my [ESPHost_EMAC](https://github.com/JAndrassy/ESPHost-EMAC/blob/main/README.md) library with a Nano Connect, which may behave differently at low level.

iabdalkader:
> For WEP different connect parameters are required and the API should be like this

I implemented the missing `WiFi.begin(ssid, keyIndex, key);` that accepts a key index. However, I don't see any connect function that accepts a key index, not in the mbed `WiFiInterface` class or in the underlying WHD STA interface, so the key index is not used. This lets you connect to WEP networks, the other option is we don't support WEP, I'm fine either way, but I don't see a reason why we must use the key index.

> If I remember it right, AUTO encryption type at connect worked, so the encryption type parameter in begin was maybe not necessary

There's no AUTO encryption type, it's just an enum we've added, maybe it maps to something valid for ESP but it does not map to anything valid for WHD. Those are the valid security types in mbedos: https://os.mbed.com/docs/mbed-os/v6.16/mbed-os-api-doxy/group__netsocket.html#ga0bfcdf6a9abae30715b5e7f43ae4a0c5

And those are the actual security types they map to in the WHD driver (use on the Giga, Nicla, Portenta): https://github.com/ARMmbed/mbed-os/blob/7ae592dabe59e9f26a4ce56190af61fb8b46d10d/connectivity/drivers/emac/COMPONENT_WHD/interface/WhdSTAInterface.cpp#L109

I tried unknown, the comments say you shouldn't but I did anyway and it does not work. So the encryption type must be specified when connecting. By default WPA/WPA2 will be used, if the network was network was not found in a scan. This should work for most cases, but it still gives the user the option to override if it does not.

JAndrassy:
sorry it was NSAPI_SECURITY_UNKNOWN not AUTO

here is how WEP is handled in WHD: https://github.com/ARMmbed/mbed-os/blob/7ae592dabe59e9f26a4ce56190af61fb8b46d10d/connectivity/drivers/emac/COMPONENT_WHD/interface/WhdSTAInterface.cpp#L329

iabdalkader:
> here is how WEP is handled in WHD: https://github.com/ARMmbed/mbed-os/blob/7ae592dabe59e9f26a4ce56190af61fb8b46d10d/connectivity/drivers/emac/COMPONENT_WHD/interface/WhdSTAInterface.cpp#L329

It seems to expect a very specific key format, I'll take a look later. Although I'm not sure if WEP is worth the trouble, it's as secure as having no security at all.

iabdalkader:
@JAndrassy I've formatted the WEP key in the same way that the driver expects, see: https://github.com/ARMmbed/mbed-os/blob/7ae592dabe59e9f26a4ce56190af61fb8b46d10d/connectivity/drivers/wifi/COMPONENT_WHD/wifi-host-driver/src/whd_wifi_api.c#L1115
However I don't have anything to test this with, there's absolutely nothing that I have that still creates a WEP AP. Also if this same code is meant to be used with other network interfaces, I'm not sure how this can be made portable.

facchinm (approved):

iabdalkader:
Okay I'll implement the OPEN `begin()` and remove WEP, and maybe leave the implementation here for reference.

iabdalkader:
Updated, `begin(ssid)` is now implemented and WEP is deprecated. 

For reference, this is the WEP key function. Since all 4 keys are required, the key is repeated.
```C++
int arduino::WiFiClass::begin(const char* ssid, uint8_t key_idx, const char* key) {
  // The low-level driver expects all 4 keys to be passed in a buffer with the following format:
  // <idx> <len> <key>, <idx> <len> <key>, etc..
  uint8_t buf[(2 + 32) * 4] = { 0 };
  size_t keylen = min(32, strlen(key));
  size_t buflen = (keylen + 2) * 4;

  // Repeat the key.
  for (int i=0; i<buflen; i += (keylen + 2)) {
      buf[i+0] = i / (keylen + 2);
      buf[i+1] = keylen;
      memcpy(&buf[i+2], key, keylen);
  }

  return begin(ssid, (const char *) buf, ENC_TYPE_WEP);
}
```

Issue #743, opened 2023-10-24
GIGA Display - Graphics offset when running on M4.


AndrewCapon:
Hi Guys,

There seems to be an issue with any code running on the M4 talking to the display.

I first noticed it with LVGL running on the M4, but it seems to effect everything.

I have enabled the pattern generator in dsi.cpp with `HAL_DSI_PatternGeneratorStart(&dsi, 1, 0);` and everything looks good here, the pattern is fine.

You can replicate with:
```
#include "Arduino_GigaDisplay_GFX.h"
#include "RPC.h"

GigaDisplay_GFX display;

void setup() 
{
  RPC.begin();
}

void loop() 
{
  display.fillScreen(0);
  display.fillRect(0,0,100,100, 0xffff);
}
```

![IMG_1500](https://github.com/arduino/ArduinoCore-mbed/assets/2297877/819861ca-c4bc-4deb-ac3b-212e150efe31)

AndrewCapon:
Ok, after a bit of investigation its the PLL that is causing the issue:

```
int dsi_init(uint8_t bus, struct edid *edid, struct display_timing *dt) {
#ifdef ARDUINO_GIGA
	static const uint32_t DSI_PLLNDIV = 125;
	static const uint32_t DSI_PLLIDF = DSI_PLL_IN_DIV3;
	static const uint32_t DSI_PLLODF = DSI_PLL_OUT_DIV1;
	static const uint32_t DSI_TXEXCAPECLOCKDIV = 4;
#else
	static const uint32_t DSI_PLLNDIV = 40;
	static const uint32_t DSI_PLLIDF = DSI_PLL_IN_DIV2;
	static const uint32_t DSI_PLLODF = DSI_PLL_OUT_DIV1;
	static const uint32_t DSI_TXEXCAPECLOCKDIV = 4;
#endif
```

On the M4 if the second set of values are used it works fine.

AndrewCapon:
so this sorts it out for both M7 and M4, just incase someone finds this before it is fixed properly:

```
int dsi_init(uint8_t bus, struct edid *edid, struct display_timing *dt) {
#ifdef ARDUINO_GIGA
	static uint32_t DSI_PLLNDIV = 125;
	static uint32_t DSI_PLLIDF = DSI_PLL_IN_DIV3;
	static uint32_t DSI_PLLODF = DSI_PLL_OUT_DIV1;
	static uint32_t DSI_TXEXCAPECLOCKDIV = 4;
[...]

PR #879, opened 2024-05-10
libraries/Arduino_H7_Video: Fix the DSI PLL configuration.

iabdalkader:
With the current configuration (NDIV=125, PLLIDF=3, PLLODF=1) the DSI PLL outputs an out of spec 83MHz clock. This fix sets the output clock to the max supported clock of 62.5MHz, according to the datasheet.

Also happens to fix issue #743

Current settings:
image


facchinm:
Well spotted! This has been wrong for quite a while, since the first alpha batch of Gigas with 12MHz quartz :man_facepalming: Merging!

PR #897, opened 2024-06-17
properly initialized and checked Ethernet init callback (fix #890)

### LINKED ISSUE #890: potential null pointer access / _initializerCallback not initialized
https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L8

When user creates his own object and passes a null pointer for EthernetInterface()
`arduino::EthernetClass Ethernet(nullptr);` then the uninitialized function pointer `_initializerCallback` is called accessing some random memory location.

Also as the comment asks "What is this function for?". If this is not defined or documentated, then you should remove it completely to avoid such errors. Beside of this, this "callback" functions **makes no sence**, as this is called from `EthernetClass::begin()` synchronously. So the user can call any function afterwards. It is **bad design** to move private project specific code into public libraries.

maidnl:
fix for potential null pointer access / _initializerCallback not initialized #890 
This PR properly initializes _initializerCallback function pointer and checks for its validity before to call it

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -5,7 +5,10 @@
 int arduino::EthernetClass::begin(uint8_t *mac, unsigned long timeout, unsigned long responseTimeout) {
   if (eth_if == nullptr) {
     //Q: What is the callback for?
-    _initializerCallback();
+    //A: To call a specific function on your system before any initialization is performed
+    if(_initializerCallback != nullptr) {
+      _initializerCallback();
+    }
     if (eth_if == nullptr) return 0;
   }
   eth_if->set_dhcp(true);
--- a/libraries/Ethernet/src/Ethernet.h
+++ b/libraries/Ethernet/src/Ethernet.h
@@ -121,7 +121,7 @@ private:
   volatile EthernetLinkStatus _currentNetworkStatus = Unknown;
   EthernetInterface net;
   EthernetInterface *eth_if = &net;
-  voidPrtFuncPtr _initializerCallback;
+  voidPrtFuncPtr _initializerCallback = nullptr;
   arduino::IPAddress ipAddressFromSocketAddress(SocketAddress socketAddress);
 };
 


libraries/Ethernet/src/Ethernet.cpp
```diff
@@ -5,7 +5,10 @@
 int arduino::EthernetClass::begin(uint8_t *mac, unsigned long timeout, unsigned long responseTimeout) {
   if (eth_if == nullptr) {
     //Q: What is the callback for?
-    _initializerCallback();
+    //A: To call a specific function on your system before any initialization is performed
```
pennam, line 8: :heart:

manchoz:
@maidnl @pennam see also https://github.com/arduino/ArduinoCore-mbed/compare/main...manchoz:ArduinoCore-mbed:rm_initCallback

maidnl:
Thanks @manchoz, however your PR removes completely the callback and this would brake backward compatibility (in case someone used the callback), so I would prefer to keep the callback and just make it safe.
What do you think @pennam ?

manchoz:
@pennam @maidnl, there is no code referring to this callback. AFAIK, it is a leftover of the early stages of the ArduinoCore-mbed. @facchinm WDYT?

pennam:
isn't already impossible to use the `_initializerCallback()` ? `eth_if` would never be a `nullptr` because is always initialized here:

https://github.com/arduino/ArduinoCore-mbed/blob/0108dedefbd748729439eee29654f49f47cf9de4/libraries/Ethernet/src/Ethernet.h#L123

maidnl:
@pennam, I could be wrong, but here is my understanding:
you have 3 constructors of EthernetClass (1) `EthernetClass(EthernetInterface *_if)`, (2) `EthernetClass();`, (3) `EthernetClass(voidPrtFuncPtr _cb)`. `_initializerCallback()` (which is pointer to function) is initialized using constructor (3) and is not related to eth_if. I agree however on the fact that eth_if could not be nullptr ever, but just because you cannot call the constructor EthernetClass(nullptr) because of the conflict between (1) and (3). 
This PR is needed in all cases since _initializeCallback is not properly initialized nor there is a check for its validity when is called.

manchoz:
Let's go ahead and remove it. I don't see any use case for this feature.

pennam:
:+1: let's go for https://github.com/arduino/ArduinoCore-mbed/pull/900

pennam:
superseded by #901

Issue #889, opened 2024-05-27
missing null pointer checks


ddmesh:
https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L6

Per default an object is created via `arduino::EthernetClass Ethernet;`. In this case the `eth_if` member got initialized (which is a pointer).

But when user creates his own object and passes a null pointer for EthernetInterface()
`arduino::EthernetClass Ethernet(nullptr);`
Several calls to other public functions, **will directly call/access the null pointer**.

https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L17
https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L56
https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L71
https://github.com/arduino/ArduinoCore-mbed/blob/dd5abb6171063b2777620b9339510c63b915ce72/libraries/Ethernet/src/Ethernet.cpp#L80

manchoz:
Hey, @ddmesh, would you mind adding the null-pointer checks where needed and submitting a PR?

Thanks!

maidnl:
@ddmesh thank you for your interest in reporting potential problems, however the scenario you described is impossible to achieve. If you try to compile something with `arduino::EthernetClass Ethernet(nullptr);` you'll get a compiler error, this means that eth_if is always initialized (`EthernetInterface *eth_if = &net;`). However I agree that is a bit confusing so I made a PR to properly check if eth_if is nullptr, just for clearness.

PR #898, opened 2024-06-17
Opta Ethernet: add nullptr check

maidnl:
This PR just add a few nullprt checks (which is always a good practice), however the scenario described in #889 (the reason for this PR) is impossible to achieve


JAndrassy:
please use at least the name of the library in the PR title. the repo has the whole platform

pennam:
superseded by #901

PR #900, opened 2024-06-18
Remove initCallback

pennam:


JAndrassy:
maybe the other two constructor could be moved above the comment for `begin`

pennam:
Superseded by #901

PR #901, opened 2024-06-18
Ethernet, initCallback, ctor and eth_if

### LINKED ISSUE #889: missing null pointer checks

### LINKED ISSUE #890: potential null pointer access / _initializerCallback not initialized

pennam:
Possible alternative to #897 #898 #900 it adds function pointers cleanup and uses the same ctor logic of WiFi.

Opened as a draft: it should build but is untested

Fixes #890 
Fixes #889

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -1,7 +1,5 @@
 #include "Ethernet.h"
 
-#define SSID_MAX_LENGTH 32
-
 int arduino::EthernetClass::begin(uint8_t *mac, unsigned long timeout, unsigned long responseTimeout) {
   if (eth_if == nullptr) {
     return 0;
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -3,11 +3,6 @@
 #define SSID_MAX_LENGTH 32
 
 int arduino::EthernetClass::begin(uint8_t *mac, unsigned long timeout, unsigned long responseTimeout) {
-  if (eth_if == nullptr) {
-    //Q: What is the callback for?
-    _initializerCallback();
-    if (eth_if == nullptr) return 0;
-  }
   eth_if->set_dhcp(true);
   return _begin(mac, timeout, responseTimeout);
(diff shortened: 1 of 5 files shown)


JAndrassy:
the remaining Ethernet constructor could be moved about the comment which belongs to `begin`

Issue #891, opened 2024-06-01
[BUG] DHCP Ethernet crashes after repeated switching of network interface [OPTA/Portenta H7]


Channel59:
I experience some issues with the network interface of the OPTA (probably also the portenta H7 suffers from this problem).

### Setup:
PC with multiple eth interface, Opta **directly connected** to `eth0`. The PC has a static IP and runs a DHCP server.  ArduinoCore-mbed version: 4.0.10

### Issue:
If from the PC the interface `eth0` is brought up and down multiple times, the OPTA sees this and the LAN light turn off when the interface is down and turn back on when the interface is up.
If the OPTA is configured to use a static IP this gives no issues. However, when DHCP is enabled, the OPTA seems to be unable to reconnect to the network. It cannot be pinged either. Also, **the network stack of the OPTA does not seem to notice this**, because the state will actually become `NSAPI_STATUS_GLOBAL_UP`, but it does not respond to pings.

_With a static IP this seems not to be an issue._

### Recovering from this state
It seems that the only way to recover from this state is by resetting the OPTA, which is undesirable. 

### Code on the OPTA:
This issue happened when using [Arduino_ConnectionHandler](https://github.com/arduino-libraries/Arduino_ConnectionHandler), just the PortentaEthernet class from the arduinocore, and it also happen when I just use the mbed EthernetInterface. 

The simplest way to reproduce the problem is with this code:
```
#include <Arduino.h>
#include "EthernetInterface.h"

auto eth = EthernetInterface::get_default_instance();

void status_callback(nsapi_event_t status, intptr_t param) ;

void setup() {
  eth->set_blocking(true);
  eth->attach(&status_callback);
  eth->set_dhcp(true);
  eth->connect();
  current_status = eth->get_connection_status();
  eth->set_blocking(false);  
}


void status_callback(nsapi_event_t status, intptr_t param) {
    Serial.println("Connection status changed!");
    switch (param) {
        case NSAPI_STATUS_LOCAL_UP:
            Serial.println("Local IP address set!");
            break;
        case NSAPI_STATUS_GLOBAL_UP:
            Serial.println("Global IP address set!");
            break;
        case NSAPI_STATUS_DISCONNECTED:
            Serial.println("No connection to network!");
            break;
        case NSAPI_STATUS_CONNECTING:
            Serial.println("Connecting to network!");
            break;
        default:
            Serial.println("Not supported");
            break;
    }
}
```

The script to break the opta:
```
[...]

PR #902, opened 2024-06-20
Ethernet OPTA: fix for dhcp not working when netif goes down and up

maidnl:
fix for dhcp not asking again for address when netif goes down and up again
this fix:
DHCP Ethernet crashes after repeated switching of network interface [OPTA/Portenta H7] #891

LWIPInterface: stop dhcp when netif goes down


maidnl:
supersede by f97304bffff1852de6ebf10ccbb941f431b70b16

PR #903, opened 2024-06-20
Ethernet OPTA: fix crash when DHCP is used on fast DHCP server

maidnl:
In some cases, when using a DHCP on an embedded device (like portenta X8) it has been observed an OPTA crash.
This patch keep the Ethernet PHY reset for a little at the beginning so that Ethernet irqs start "arriving" when all is properly set -up.

OPTA: keep ethernet phy under reset at boot

### DIFF
--- a/variants/OPTA/pins_arduino.h
+++ b/variants/OPTA/pins_arduino.h
@@ -193,6 +193,8 @@ void _ontouch1200bps_();
 
 #define USB_MAX_POWER	(500)
 
+#define ETHERNET_PHY_RESET   (185)
+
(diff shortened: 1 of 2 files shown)


pennam:
@maidnl this causes a red blink of death on my OPTA. I can't reproduce the issue using core 4.1.3, but with this patch on top of https://github.com/arduino/ArduinoCore-mbed/commit/546529afa24a2add5b566e1c4c0c9d772a424c25 i can easily reproduce the issue just unplugging and replugging the usb cable.

maidnl:
This pull request is not needed. Found the root cause of the crashes elsewhere.

PR #907, opened 2024-06-27
MbedClient add network status check

pennam:
This PR makes `connected()`  consider also network status. In some conditions: WiFi turning ON and OFF or Ethernet cable connections and disconnections `connected()` returns 1 even if network is in reality disconnected.

~~Also `status()` is a bit confusing so i make it private, users can still check Client status using `connected()`~~


JAndrassy:
please let status public. `connected` returns true if data are available, which is not useful for sending.
in the W5500 Ethernet library status() even allows async connect if the connect timeout is low or zero.

https://github.com/Networking-for-Arduino/Arduino-Networking-API/blob/main/ArduinoNetAPIDev.md#status-optional

PR #912, opened 2024-06-29
SocketWrapper MbedClient debugging readSocket

JAndrassy:
1) In `connect` we have to delete the socket object, because all other functions test it for null and then use mutex which is null, because configureSocket was not invoked. And there are too few sockets in total.

2) In `readSocket` the condition of the inner loop was always true and the inner loop never exited to wait for event or timeout on event. The inner loop only paused for yield(). With the PR if no data are available the inner loop does `break` to outer loop where the thread waits for event or timeout. 

<del>3) I think we have to do mutex in `write` too</del>

### DIFF
--- a/libraries/SocketWrapper/src/MbedClient.cpp
+++ b/libraries/SocketWrapper/src/MbedClient.cpp
@@ -22,28 +22,30 @@ void arduino::MbedClient::readSocket() {
     int ret = NSAPI_ERROR_WOULD_BLOCK;
     do {
       mutex->lock();
-      if (sock != nullptr && rxBuffer.availableForStore() == 0) {
+      if (sock == nullptr) {
+        mutex->unlock();
+        goto cleanup;
+      }
+      if (rxBuffer.availableForStore() == 0) {
         mutex->unlock();
         yield();
         continue;
-      } else if (sock == nullptr) {
-        goto cleanup;
       }
       ret = sock->recv(data, rxBuffer.availableForStore());
       if (ret < 0 && ret != NSAPI_ERROR_WOULD_BLOCK) {
+        mutex->unlock();
         goto cleanup;
       }
       if (ret == NSAPI_ERROR_WOULD_BLOCK || ret == 0) {
-        yield();
         mutex->unlock();
-        continue;
+        break;
       }
       for (int i = 0; i < ret; i++) {
         rxBuffer.store_char(data[i]);
       }
       mutex->unlock();
       _status = true;
-    } while (ret == NSAPI_ERROR_WOULD_BLOCK || ret > 0);
+    } while (true);
   }
 cleanup:
   _status = false;
@@ -98,6 +100,7 @@ int arduino::MbedClient::connect(SocketAddress socketAddress) {
   }
 
   if (static_cast<TCPSocket *>(sock)->open(getNetwork()) != NSAPI_ERROR_OK) {
+    _status = false;
     return 0;
   }
 
@@ -117,6 +120,7 @@ int arduino::MbedClient::connect(SocketAddress socketAddress) {
     configureSocket(sock);
     _status = true;
   } else {
+    sock->close();
     _status = false;
   }
 
@@ -148,6 +152,7 @@ int arduino::MbedClient::connectSSL(SocketAddress socketAddress) {
   }
 
   if (static_cast<TLSSocket *>(sock)->open(getNetwork()) != NSAPI_ERROR_OK) {
+    _status = false;
     return 0;
   }
 
@@ -179,6 +184,7 @@ restart_connect:
     configureSocket(sock);
     _status = true;
   } else {
+    sock->close();
     _status = false;
   }
 
@@ -224,8 +230,9 @@ int arduino::MbedClient::available() {
 }
 
 int arduino::MbedClient::read() {
-  if (sock == nullptr)
+  if (mutex == nullptr) {
     return -1;
+  }
   mutex->lock();
   if (!available()) {
     mutex->unlock();
@@ -238,8 +245,9 @@ int arduino::MbedClient::read() {
 }
 
 int arduino::MbedClient::read(uint8_t *data, size_t len) {
-  if (sock == nullptr)
+  if (mutex == nullptr) {
     return 0;
+  }
   mutex->lock();
   int avail = available();
 
@@ -261,7 +269,14 @@ int arduino::MbedClient::read(uint8_t *data, size_t len) {
 }
 
 int arduino::MbedClient::peek() {
-  return rxBuffer.peek();
+  if (mutex == nullptr) {
+    return 0;
+  }
+  mutex->lock();
+  int res = rxBuffer.peek();
+  mutex->unlock();
+
+  return res;
 }
 
 void arduino::MbedClient::flush() {


andreagilardoni:
Hi @JAndrassy, I agree with the changes you made for point 1.
For point 2 are you trying to fix a particular issue that is affecting you? Apart from missing mutex locks and unlocks and the nice code reordering, I don't see any huge changes in code that may impact functionality on `MbedClient`.
For point 3, I expect it to work without the need of mutex, since I think tx and rx buffer are separated, I will take a look at the lower level implementation in order to be sure

JAndrassy:
@andreagilardoni so should we add everywhere to `if (sock == NULL)` `&& `mutex == NULL`?

https://os.mbed.com/docs/mbed-os/v6.16/mbed-os-api-doxy/class_socket.html#abc36eff670a3bee145f1c629a7eb7ee3
has
> If connect() fails it is recommended to close the Socket and create a new one before attempting to reconnect.

I didn't look in the source but I think the unconnected socket still counts as used one from socket-max

andreagilardoni:
I think the only method missing the check for ``sock==nullptr`` is read(), I think it is enough to return -1 there. I think it is just enough to check for `sock`, since mutex should be != nullptr when the sock is != nullptr after the changes you proposed here.

The changes on lines 123:124 are necessary. 
The changes on line 220, 224 I don't think that may affect anything except performances. 

I only need to understand why you performed changes on `readSocket()`, since to me everything seems to be a rearrangement on if statements (with some additional actions on mutex). Did you experience any kind of issue that I can try to replicate?

JAndrassy:
> changes on readSocket()

`continue` change to `brreak` if no data are available to read

libraries/SocketWrapper/src/MbedClient.cpp
```diff
@@ -22,28 +22,31 @@ void arduino::MbedClient::readSocket() {
...
       mutex->lock();
-      if (sock != nullptr && rxBuffer.availableForStore() == 0) {
+      if (sock == nullptr) {
+        mutex->unlock();
+        goto cleanup;
+      }
+      if (rxBuffer.availableForStore() == 0) {
         mutex->unlock();
         yield();
         continue;
```
andreagilardoni, line 32:
```suggestion
        break;
```

JAndrassy:
changes:
* for unsuccessful connect `socket->close` (it was `delete sock` in the first version of the PR)
* make sure mutex is not used when it is null

libraries/SocketWrapper/src/MbedClient.cpp
```diff
@@ -224,7 +231,7 @@ int arduino::MbedClient::available() {
 }
 
 int arduino::MbedClient::read() {
-  if (sock == nullptr)
+  if (mutex == nullptr)
     return -1;
```
andreagilardoni, line 235:
```suggestion
  if (mutex == nullptr) {
    return -1;
  }
```

libraries/SocketWrapper/src/MbedClient.cpp
```diff
@@ -238,7 +245,7 @@ int arduino::MbedClient::read() {
 }
 
 int arduino::MbedClient::read(uint8_t *data, size_t len) {
-  if (sock == nullptr)
+  if (mutex == nullptr)
     return 0;
```
andreagilardoni, line 249:
```suggestion
  if (mutex == nullptr) {
    return 0;
  }
```

andreagilardoni:
@JAndrassy after looking at that everything seems ok, I need to try to run some examples and then I think we can merge this PR. I like how you reorganized the thread function, thanks for your contribution!

JAndrassy:
There is one more problem with the readSocket thread, but if I patch the solution in, it screams for a rewrite of the whole inner loop.

the problem is that on err < 0 the thread ends. it can be just that the peer closed the connection.  on a new `connect` without `stop()`, the connection is established, but the readSocket thread is not started. It is not possible to restart a thread, so it should not end until sock is null.

andreagilardoni (dismissed):

andreagilardoni:
I quickly tested this PR and I don't see any issues with this. About the other issue you are describing, what `err < 0` are you talking about?

JAndrassy:
sorry ret not err
```
      if (ret < 0 && ret != NSAPI_ERROR_WOULD_BLOCK) {
        mutex->unlock();
        goto cleanup;
      }
 ```

**Comment by andreagilardoni** [comment]
If the peer closed the connection than, I think, it is expected that the socket has to be closed and needs to be restarted. In connect I can see that this check is performed https://github.com/arduino/ArduinoCore-mbed/blob/2ece915a631f0decffacf8839133278f59d44507/libraries/SocketWrapper/src/MbedClient.cpp#L84-L89

Did I get your point?

**Comment by JAndrassy** [comment]
then it is ok

**Review by maidnl — CHANGES_REQUESTED** [review]
I tried to analyse and test the proposed changes and I have a few comments (I hope you can find them meaningful).

One thing I found really confusing is the use of the `_status` flag which is set / reset in a lot of different places (and this make difficult to understand the underling logic). 
In the `readSocket()` thread is appears that is not really necessary to set `_status=true` at the end of each innermost `do-while` cicle.
There is no need to set `_status=true` after the call of the `configureSocket()` function since is the `configureSocket()` function itself sets `_status=true;`.  This happens twice: in the `connect()` and the `connectSSL()` function. 
As additional simplification I would reset `_status` to false at the beginning of those 2 functions (if all goes well the `configureSocket()` function will set it to `true` and in case of problem it will remain set to `false` removing the need of setting if to `false` in case of errors). This removes the need to set `_status` to false in case of problem and ensure it is false if something wrong happened.

About the mutex lock/unlock in the `write()` function, my understanding is that the mutex is used to prevent access to the `rxBuffer` from different threads, so it appears to be not necessary here. One point that is certainly wrong is that the changes removed the check about the `sock` pointer: it is necessary to reintroduce here the check 
```  
if (sock == nullptr) {
   return 0;
}
```
this is very important because a user can call `write()` with an invalid `sock` and this would crash the program.

Since the mutex is used to prevent "common" access to `rxBuffer` I noticed that the `peek()` function that uses `rxBuffer` is unprotected: I would add the  mutex lock/unlock mechanism to the `peek()` function.

A possible improvement is related to the `setSocket()` function: this function is called when client is create by a server. However the server allocates a `Socket` only if there is an incoming request, in case no request is made the server will always call `setSocket(nullptr)` (please check the `EthernetServer::available()` function in the EthernetServer.cpp file and verify if my understanding is correct). 
In case like this it is probably pointless to call `configureSocket()` so I would add a check and execute the body of the function only  if `_sock` is different from nullptr.

My last remark is more a doubt: in the `connect()` and `connectSSL()` function almost at the end of the function in case `ret` is not 1 it has been added the statement `sock->close()`, but this only closes the `Socket`. Would not be better to call directly the `MbedClient` `stop()` function? This would reset all the variables held by the client and not only "close the socket".

**Comment by JAndrassy** [comment]
`_status` only exists for `status()`. there is no other simple way.  It has no internal use in MbedClient. This PR is not about _status.

I avoid doing unnecessary changes in my PR. Where would I stop if I begin to cleanup the code. So I don't even start. So the superfluous `_status=true` stay there for this PR.

The readSocket() function runs in a separate thread so it can't switch _status to false if it isn't definitive.

 I can remove the lock from write. (btw if socket is null, then mutex is null too)

yes. peek() needs the lock. I add it.

> My last remark is more a doubt: in the connect() and connectSSL() function almost at the end of the function in case ret is not 1 it has been added the statement sock->close(), but this only closes the Socket. Would not be better to call directly the MbedClient stop() function? This would reset all the variables held by the client and not only "close the socket".

as I understand it, the idea is that the socket can be reused for next try to connect, that is why it is not deleted. calling stop() would delete it. all other fields are not initialized because configureSocket didn't run

**Comment by JAndrassy** [comment]
> I can remove the lock from write. (btw if socket is null, then mutex is null too)
> yes. peek() needs the lock. I add it.

I made these changes ^^^

**Comment by facchinm** [comment]
Maybe this patch could help fixing  #937 ?

File: libraries/SocketWrapper/src/MbedClient.cpp [hunk 4]
```diff
@@ -22,28 +22,30 @@ void arduino::MbedClient::readSocket() {
     int ret = NSAPI_ERROR_WOULD_BLOCK;
     do {
       mutex->lock();
-      if (sock != nullptr && rxBuffer.availableForStore() == 0) {
+      if (sock == nullptr) {
```
schnoberts1, line 25: I know I’ve mentioned this before but heavy optimisation on a very modern compiler could break this and it’ll be hard to tell. The compiler can’t tell this is in a thread and sock isn’t marked in anyway to inform the compiler  that it needs to read sock out of memory.

ppp-one:
Just throwing #1067 here. Queries to the Arduino Opta hang for ~100ms due to `event->wait_any(0xFF, 100);`

PR #915, opened 2024-07-01
SocketWrapper lib Client setConnectionTimeout added

JAndrassy:
In https://github.com/arduino/ArduinoCore-mbed/pull/769 I wrote in a comment

> it is not a connection timeout as setConnectionTimeout should do! it is timeout on other operations, not on connect.

reading the Mbed TCPSocket.cpp source, I see that this is not true and Mbed uses that timeout in `connect`. Sorry. My mistake was my assumption that it works the same way as in LwIP, but Mbed adds some logic as it wraps LwIP.

this PR adds setConnectionTimeout as alias to setSocketTimeout and sets the timeout before `connect`

this timeout is now applied in MbedClient in `connect`, `stop` and `write`. `read` shouldn't block in Arduino API so in `read` the timeout is not used.

### DIFF
--- a/libraries/SocketWrapper/src/AClient.cpp
+++ b/libraries/SocketWrapper/src/AClient.cpp
@@ -18,6 +18,11 @@ void arduino::AClient::setSocket(Socket *sock) {
   client->setSocket(sock);
 }
 
+void arduino::AClient::setConnectionTimeout(unsigned long timeout) {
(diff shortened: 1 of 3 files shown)


JAndrassy:
closing, because the timeout is 30 seconds in lwIP so what Mbed's TCPSocket does in its `connect` doesn't do anything in case of lwIP as net stack

PR #921, opened 2024-07-09
Add support for DHCP hostname

Channel59:
Recently someone pointed me to this PR on the mbed-os repo: https://github.com/ARMmbed/mbed-os/pull/15506/. 
The patches to make these changes are included in this pull request and so the credits for these patches full go to https://github.com/guilhermerc.

It enables the use of a hostname when DHCP is used to obtain an IP address.

Subsequently, the Ethernet.cpp and Ethernet.h are adjusted to allow the hostname to be set:

~~Either using `Ethernet.begin("desired-hostname");`~~

or using a somewhat more Arduino-like api:
```cpp
Ethernet.setHostname("desired-hostname");
Ethernet.begin();
```

Before this compiles, `libmbed.a` should be recompiled on https://github.com/ARMmbed/mbed-os/tree/latest which will also patch some of the header files that are required to compile the arduino program.

The code seems to work correctly with the opta. Any testing / refinements are more than welcome.

I do not have a wifi enabled device, so I cannot test it, but with these patches applied, a small change to Wifi.h and Wifi.cpp could add support for Wifi as well.


Channel59:
The compiler errors arise from the fact that a call to Ethernet.begin(nullptr) is ambiguous now.
I suppose to make sure no code is broken the `Ethernet.begin("desired-hostname");` api should be removed and only the setHostname should be used.

If someone has a more elegant solution please let me know.

pennam:
Hi @Channel59 thank you so much for your contribution!
I took your commits and squashed a bit, also fixed patch numbers and added support for wifi. If everithing works and you are ok with this i will close this PR and merge this https://github.com/arduino/ArduinoCore-mbed/pull/969

Channel59:
Hi @pennam, The addition of the wifi support is great! I suppose this PR can be closed and #969 should be merged instead.

PR #934, opened 2024-08-01
RPC/OpenAMP: Support loading sketches to SDRAM.

iabdalkader:
To use this with MicroPython, to load Arduino sketches to SDRAM, it requires https://github.com/micropython/micropython/pull/15499
Which has been merged, so a nightly build will work.

### DIFF
--- a/libraries/RPC/src/RPC.cpp
+++ b/libraries/RPC/src/RPC.cpp
@@ -1,3 +1,5 @@
+// Copyright (c) 2024 Arduino SA
+// SPDX-License-Identifier: MPL-2.0
 #include "RPC.h"
 
 #define ENDPOINT_ID_RAW     0
--- a/libraries/RPC/src/RPC.cpp
+++ b/libraries/RPC/src/RPC.cpp
@@ -163,6 +163,35 @@ int RPCClass::begin() {
 #endif
 
 #ifdef CORE_CM4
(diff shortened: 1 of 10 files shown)


libraries/RPC/src/RPC.cpp
```diff
@@ -163,6 +163,35 @@ int RPCClass::begin() {
...
+        MPU_InitStruct.IsShareable = MPU_ACCESS_NOT_SHAREABLE;
+        MPU_InitStruct.IsCacheable = MPU_ACCESS_NOT_CACHEABLE;
+        MPU_InitStruct.IsBufferable = MPU_ACCESS_NOT_BUFFERABLE;
+        MPU_InitStruct.SubRegionDisable = 0x00;
+        HAL_MPU_ConfigRegion(&MPU_InitStruct);
+        HAL_MPU_Enable(MPU_PRIVILEGED_DEFAULT);
+    }
+};
+
+M4Init dt __attribute__ ((init_priority (101)));
```
facchinm, line 192: I would change the name of the object being created (something like `__m4init` instead than `dt`)

facchinm (changes requested):

libraries/RPC/src/RPC.cpp
iabdalkader: Done!

PR #948, opened 2024-09-02
PDM: fix setGain() for dfsdm

facchinm:
If the gain was too high, the right bit shift
  `RecBuff[i] >> attenuation`
would result in undefined behaviour
https://www.iso-9899.info/n1570.html#6.5.7

Take the chance to also fix the (approximate) conversion from gain (in dB) to bit shift


d-le0:
All good!
Also the behavior of setGain is intuitive.

PR #957, opened 2024-09-23
[STM32H7] Provisioning fs reformat

andreagilardoni:
Changing flash partitioning scheme in order to make room for a KV store

### DIFF
--- a/libraries/STM32H747_System/examples/QSPIFormat/QSPIFormat.ino
+++ b/libraries/STM32H747_System/examples/QSPIFormat/QSPIFormat.ino
@@ -83,14 +83,29 @@ void setup() {
       // use space from 15.5MB to 16 MB for another fw, memory mapped
     }
 
-    int err = wifi_data_fs.reformat(&wifi_data);
-    if (err) {
+    bool reformat = true;
+
+    if(!wifi_data_fs.mount(&wifi_data)) {
+      Serial.println("\nPartition 1 already contains a filesystem, do you want to reformat it? Y/[n]");
+      wifi_data_fs.unmount();
+
+      reformat = waitResponse();
+    }
+
+    if (reformat && wifi_data_fs.reformat(&wifi_data)) {
       Serial.println("Error formatting WiFi partition");
       return;
     }
 
-    err = ota_data_fs.reformat(&ota_data);
-    if (err) {
+    reformat = true;
+    if(!ota_data_fs.mount(&ota_data)) {
+      Serial.println("\nPartition 2 already contains a filesystem, do you want to reformat it? Y/[n]");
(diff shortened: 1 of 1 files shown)


pennam (changes requested):
There is no need to erase the whole flash before repartitioning; you only need to erase the MBR.

root.init();
root.erase(0x0, root.get_erase_size());

Furthermore if you keep the KVstorage in partition 3 for both schemes you should be able switch between them without erase the MBR at all


```
mbed::MBRBlockDevice::partition(&root, 1, 0x0B, 0, 1024 * 1024);
if(default_scheme) {
  mbed::MBRBlockDevice::partition(&root, 4, 0x0B, 14 * 1024 * 1024, 14 * 1024 * 1024);
  mbed::MBRBlockDevice::partition(&root, 3, 0x0B, 13 * 1024 * 1024, 14 * 1024 * 1024);
  mbed::MBRBlockDevice::partition(&root, 2, 0x0B, 1024 * 1024, 13 * 1024 * 1024);
  // use space from 15.5MB to 16 MB for another fw, memory mapped
} else {
  mbed::MBRBlockDevice::partition(&root, 2, 0x0B, 1024 * 1024, 6 * 1024 * 1024);
  mbed::MBRBlockDevice::partition(&root, 3, 0x0B, 6 * 1024 * 1024, 7 * 1024 * 1024);
  mbed::MBRBlockDevice::partition(&root, 4, 0x0B, 7 * 1024 * 1024, 14 * 1024 * 1024);
  // use space from 15.5MB to 16 MB for another fw, memory mapped
}
```

andreagilardoni:
> Furthermore if you keep the KVstorage in partition 3 for both schemes you should be able switch between them without erase the MBR at all

I tried switching from this scheme to the older one and something bad happened and the only way to make it work properly was to erase the partitions. I am double checking this.

andreagilardoni:
I confirm that without the MBR erase I cannot use the old partitioning scheme.
@pennam do you think it is better to split this PR into 2? the first one with just the MBR erase command and the other with the new partitioning scheme, so that one can rollback with this fix present.

megacct:
@andreagilardoni @pennam Hi, a couple of us have tried, unsuccessfully, to scrub the MBR in order to change the partition layout. Detail here https://forum.arduino.cc/t/how-to-partition-giga-flash-similar-to-original-factory-defaults/1349759. I'm sure we were doing something wrong but it may be worth double-double-checking that the erase in this commit is doing what's expected.

andreagilardoni:
I would like to point out that when repartitioning, for the user or ota block devices some files can be corrupted if they lie in the area that will be used for partition 4.

pennam:
superseded by #1058

PR #962, opened 2024-09-26
Reduce the number of WiFi join retry to avoid watchdog reset

pennam:
Maximum watchdog window is 32s, join worst case 7 * 5 = 35s

This PR reduces the max number of retries to 1 to block at most 7 seconds


pennam:
Note to self: remember to update patch number before merge :slightly_smiling_face:

Issue #965, opened 2024-10-01
GIGA R1: kicking the dog causes crash


megacct:
if the watchdog hasn't been started. This is contrary to the comments here. Easily overcome by using is_running() first.

https://github.com/arduino/ArduinoCore-mbed/blob/931f0bbe857851d51df0e1814ab25cb5cf56c9e2/cores/arduino/mbed/drivers/include/drivers/Watchdog.h#L144

PR #967, opened 2024-10-02
patch: kik watchdog only if running

pennam:
fixes: #965


andreagilardoni:
I verified this PR with the following code:
```cpp
#include <Watchdog.h>

void setup() {
  mbed::Watchdog::get_instance().kick();
}
```

andreagilardoni (approved):

PR #969, opened 2024-10-03
WiFi and Eth Dhcp hostname support

pennam:
following https://github.com/arduino/ArduinoCore-mbed/pull/921
Tested on Nicla Vision Giga R1 OPTA (Eth/WiFi) Portenta (Eth/WiFi)
Merge after #983


pennam:
CI should stop complaining as soon as we rebuild libmbed.a

PR #976, opened 2024-10-21
WiFi: add setTimeout()

pennam:
OPTA :+1: 
PORTENTA H7 :+1: 
GIGA R1 :+1: 
NICLA_VISION :+1: 

ci sill stop complain as soon as a global libmbed.a rebuild is done

### DIFF
--- a/libraries/WiFi/src/WiFi.cpp
+++ b/libraries/WiFi/src/WiFi.cpp
@@ -75,6 +75,7 @@ int arduino::WiFiClass::begin(const char* ssid, const char* passphrase, wl_enc_t
     wifi_if->set_network(_ip, _netmask, _gateway);
   }
 
+  wifi_if->set_timeout(_timeout);
   nsapi_error_t result = wifi_if->connect(ssid, passphrase, _security);
 
   if(result == NSAPI_ERROR_IS_CONNECTED) {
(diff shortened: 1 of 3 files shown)


Nopathway (changes requested):
Add this changes

rishita-asl:
Hii, could you share a bit more documentation as to what the actual purpose and effect of the setTimeout function is? The comment "Configure WiFi join timeout in milliseconds. Default value is 7s" isn't self explanatory, and I am confused as about what effect this timeout value will have on an connection. What does "WiFi join" mean?

pennam:
`WiFi.begin()` is a blocking API, when you call it, the wifi driver tries to join (connect) to the configured wifi network. If for some reason  the WiFi chip can't join the network; the wifi driver was configured to do a number of [retries](https://github.com/arduino/mbed-os/commit/e959ae057b31849f779904ce97bfabc68658f53e) each of that would block execution up to [7s](https://github.com/arduino/mbed-os/blob/efd54c8990ba5b437eb4eb8b786b7e48941b03f1/connectivity/drivers/wifi/COMPONENT_WHD/wifi-host-driver/src/whd_wifi_api.c#L67).

With this [patch](https://github.com/arduino/ArduinoCore-mbed/pull/962) we have reduced the number of internal retries to 1.

For some applications you may want to reduce the join timeout even more, and this is possible using `WiFi.setTimeout()`

PR #980, opened 2024-10-25
WiFi Ethernet deprecate MACAddress(...) in favour of macAddres(...)

pennam:
`MACAddress` is deprecated keeping the original reversed behaviour.
`uint8_t* macAddress(uint8_t * mac)` returns byte aligned with `String macAddress()`


pennam:
> Ethernet API never had the MAC address reversed.

Thanks @JAndrassy  i will doublecheck it, I was assuming it was reversed also in Ethernet library

My goal was primarly to align the behaviour of `macAddress` at least between mbed and renesas core. At this point I suppose is it better to introduce the "reverse tradition" in the renesas WiFi library

JAndrassy:
> t this point I suppose is it better to introduce the "reverse tradition" in the renesas WiFi library

we removed it recently

pennam:
> we removed it recently

yeah i've disagreed with myself one second after writing this :slightly_smiling_face: 

I will doublecheck the Ethernet MACAddress behaviour , It feels a bit strange to me that is not reversing the byte order since it is using the same socketWrapper function used in WiFi macAddress but maybe it get twisted one more time before.

JAndrassy:
I meant Ethernet API in general. the Wiznet Ethernet library, the UIPEtherner etc

pennam:
ah ok, got it!

pennam:
rebased on latest main to fix ci.
 - `WiFi.MACAddress(uint8_t *mac)` deprecated since this is Ethernet API
 - `MbedSocketClass.macAddress(uint8_t *mac)` fixed reversed order

andreagilardoni (approved):

PR #986, opened 2024-11-01
networking - move setHostname to base class SocketWrapper

JAndrassy:


pennam (approved):
Thanks for the changes @JAndrassy

PR #992, opened 2024-11-08
Add support for client certificate and private key

pennam:


libraries/SocketWrapper/src/AClient.h
andreagilardoni: I would add a comment specifying that these functions has to be called before `connect`

PR #995, opened 2024-11-10
Change the FMC frequency source to HCLK

Bexin3:
Change the FMC frequency source to HCLK, freeing up PLL2 for other uses. On Portenta, the frequency will be same. On giga it will be higher but still within spec, and timings should get recalculated based on MICROPY_HW_SDRAM_FREQUENCY.


Bexin3:
Had to do a small change, should be in a final form now though.

megacct:
Applied this PR to my GIGA and ran its SDRAM intensive workload for 90mins or so. Nil impact. I will continue running with these changes and report any issues

PR #1009, opened 2024-12-30
Reduce nsapi dns response timeout:  5 seconds 3 retires

fabik111:
Currently the nsapi is configured with a default value of 10s of timeout and with 10 attempts before failure.
In case internet is not available (but the board is correctly connected to the AP), with this set of parameters the sketch is waiting for 100 s stopping the main loop. If you are using an ArduinoIoTCloud sketch the board is rebooted by the watchdog timer

### DIFF
--- a/variants/GIGA/conf/mbed_app.json
+++ b/variants/GIGA/conf/mbed_app.json
@@ -11,6 +11,8 @@
       "rtos.main-thread-stack-size": 32768,
       "cordio.max-connections": 5,
       "target.mbed_app_start": "0x8040000",
+      "nsapi.dns-response-wait-time": 5000,
+      "nsapi.dns-total-attempts": 3,
       "target.macros_add": [
         "METAL_INTERNAL",
         "VIRTIO_DRIVER_ONLY",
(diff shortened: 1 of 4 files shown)


pennam (changes requested):
@fabik111 the changes for the RP2040 and the GENERIC_STM32H7_M4 variant should be removed.

RP2040 uses WiFiNINA, and the M4 variant does not support networking.

pennam:
@fabik111 according to [this](https://www.rfc-editor.org/rfc/rfc1123.html#page-74:~:text=If%20this%20information%20is%20not%20available%2C%20a%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20default%20of%20no%20less%20than%205%20seconds%20should%20be%20used) i would feel more safe  increasing the timeout to 5s and reducing the number of retries to 3.
Also windows [dnslookup](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc788023(v=ws.11)) and linux host command are using a default timeout of 5 seconds

pennam (approved):

pennam:
Removed also edge control and squashed changes

PR #1014, opened 2025-01-08
Ping command

fabik111:
This PR adds the `Ping` command to SocketHelpers making it available from all the IP network interfaces (WiFi, Ethernet, CatM1...).

The method signature is:
```
  int ping(const char* hostname, uint8_t ttl = 128);
  int ping(const String& hostname, uint8_t ttl = 128);
  int ping(IPAddress host, uint8_t ttl = 128);
```


pennam:
Force pushed to fix conflicts and rebase on `master`. Other changes squashed in the first commit:
* Add comments about ttl
* Remove RP2040 changes in json file

Added patch with ICMPSocket ping. We need to move ping implementation inside libmbed to avoid adding lwIP include paths to the build. lwIP paths were intentionally removed to avoid clashes in non case sensitive systems.

https://github.com/arduino/ArduinoCore-mbed/commit/de1baabed875cce3977a1092d47307e966c3188b

Adapt SocketHelpers ping to use ICMPSocket ping

pennam:
| Board  | Eth |WiFi|
| ------------- | ------------- |------------- |
| OPTA | :+1:   | :+1:   |
| H7  | :+1:   | :+1:   |
| GIGA  | NA   | :+1:   |
| NV  | NA   | :+1:   |

megacct:
Hi, could timeout be added to the public interfaces as I'd like to override the default 5000

pennam:
@megacct we can try to add it in the next release. You can provide a PR if you wish.

megacct:
Thanks @pennam . Not familar with the process but will investigate

PR #1015, opened 2025-01-10
Arduino_H7_Video: handling errors at initialization time

dmazzella:
handling errors at initialization time with added retry counter to avoid infinite wait on anx7625_wait_hpd_event

### DIFF
--- a/libraries/Arduino_H7_Video/src/Arduino_H7_Video.cpp
+++ b/libraries/Arduino_H7_Video/src/Arduino_H7_Video.cpp
@@ -95,7 +95,10 @@ int Arduino_H7_Video::begin() {
 #endif
 
   /* Video controller/bridge init */
-  _shield->init(_edidMode);
+  int err_code = _shield->init(_edidMode);
+  if (err_code < 0) {
+    return err_code;
+  }
 
   #if __has_include("lvgl.h")
     /* Initiliaze LVGL library */
--- a/libraries/Arduino_H7_Video/src/Arduino_H7_Video.cpp
+++ b/libraries/Arduino_H7_Video/src/Arduino_H7_Video.cpp
@@ -97,7 +97,7 @@ int Arduino_H7_Video::begin() {
   /* Video controller/bridge init */
   int err_code = _shield->init(_edidMode);
   if (err_code < 0) {
-    return err_code;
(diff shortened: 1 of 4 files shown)


dmazzella:
the motivation for this pr is to be able to reboot the board in case the display (USBCVideo) does not turn on for some reason, unfortunately it often happens that after a reset (ota firmware update) of the board the display does not turn on again and a reset fixes this, the problem is that in my specific case it happens on “remote” machines and not being physically there to reset it this is a problem.

```c++
  // Initialize display
  int display_ec = Display.begin();
  if (display_ec != 0) {
    NVIC_SystemReset();
  }
```

dmazzella:
@leonardocavagnis please take a look at it and let me know.

libraries/Arduino_H7_Video/src/Arduino_H7_Video.cpp
```diff
@@ -95,7 +95,10 @@ int Arduino_H7_Video::begin() {
 #endif
 
   /* Video controller/bridge init */
-  _shield->init(_edidMode);
+  int err_code = _shield->init(_edidMode);
+  if (err_code < 0) {
+    return err_code;
```
leonardocavagnis, line 100:
```suggestion
    return 3; /* Video controller fail init */
```

leonardocavagnis (changes requested):
Sounds good!
Just a small fix: the `Display.begin()` function returns `0` on success or a value `>0` if it fails.

0: Success
1: Unknown error
2: Insufficient drawing memory buffer
(NEW) 3: Video controller initialization failure

iabdalkader:
Changes in this PR break the USB-C display for me (also see https://github.com/arduino/ArduinoCore-mbed/issues/1075 which might be related). I think the main issue is `anx7625_dp_get_edid` fails for some displays, and then `anx7625_dp_start` never gets called. However, because the examples don't handle `begin` failing, they proceed to call other functions normally which somehow results in a hardfault. The same goes for waiting for HPD event. If you connect the board to a PC instead, and just wait, it hard-faults after a while.
I suggest we revert some of these changes, maybe keep the `dp_start` check, and fix the examples to handle `begin` failing. Note that calling `begin` multiple times, if it fails, does not work.

PR #1016, opened 2025-01-20
Make uniform the returned WiFi FW Version value

fabik111:
Uniform the returned WiFi FW Version value with other board cores removing the first character `'v'`


pennam:
closed in favour of #1020

Issue #448, opened 2022-04-07
Raspberry Pi Pico (RP2040) - I2C holds SCL down


marmil:
Hi everybody,

i am currently trying to implement an I2C connection between an ESP32-WROVER and a Raspberry Pi Pico (RP2040). The ESP is configured as master and shall request data from the Pico which is configured as slave. Besides requesting data the master also sends data to the slave.
After encountering some issues that the I2C communication is not working properly i created a minimal example for master and slave where the problem still occurs. Regularly the I2C times out (diagnosed by ESP via Debug messages and Logic analyzer - see below) and no data can be transmitted anymore. Pullups of SDA/SCL should be fine with 3K3.

First the program-code of the slave, then some measures&measurements i already took/did and in the end the despaired beg for help finding the problem which i assume in the I2C implementation.

**Slave (Pico) Code:**

```
#include <Arduino.h>
#include <Wire.h>

#define DATABYTES   6

unsigned long lastLEDToggle = 0;
#define LED_INTERVAL 750
bool LEDstatus = HIGH;

#define I2C_SLAVE_ADDRESS 0x18

MbedI2C myI2C(14, 15);
uint8_t tmpRcvdData[DATABYTES];
uint8_t tmpRcvdDataInc;

uint8_t writeBuffer[DATABYTES] = {0x12};

void initCommunication();
void setupCommunication();

void onReceive(int numBytes);
void onRequest();


void setup() {

  pinMode(LED_BUILTIN, OUTPUT);
  
  setupCommunication();

}

void loop() {

  if (millis() - lastLEDToggle > LED_INTERVAL) {

    digitalWrite(LED_BUILTIN, LEDstatus);
    LEDstatus = !LEDstatus;

    lastLEDToggle = millis();

  }

}

void setupCommunication() {
  
  initCommunication();

  myI2C.onReceive(onReceive);
  myI2C.onRequest(onRequest);

}

void initCommunication() {

  myI2C.begin(I2C_SLAVE_ADDRESS);
  myI2C.setClock(800000);

}

void onReceive(int numBytes) {

  tmpRcvdDataInc = 0;

  while (myI2C.available() && tmpRcvdDataInc < DATABYTES) {
    tmpRcvdData[tmpRcvdDataInc] = myI2C.read();
    tmpRcvdDataInc++;
  }

  while (myI2C.available() ) {
    myI2C.read(); // clear buffer
  }
  
  
}

void onRequest() {

[...]

PR #1018, opened 2025-01-23
fix I2C slave communication stuck

maidnl:
using I2C as __slave__ leads to I2C bus stuck and no more communication possible, this seems to be the same than issue #448 
This strange behavior has been observed on Nordic nRF52840 is due to an unprotected access to rxBuffer in I2C slave implementation
This PR protects access to rxBuffer using core util critical section enter / exit function

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -122,14 +122,22 @@ size_t arduino::MbedI2C::write(const uint8_t* data, int len) {
 }
 
 int arduino::MbedI2C::read() {
+	int rv = -1;
+	core_util_critical_section_enter();
+	
 	if (rxBuffer.available()) {
-		return rxBuffer.read_char();
+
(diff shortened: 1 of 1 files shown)


libraries/Wire/Wire.cpp
```diff
@@ -122,14 +122,23 @@ size_t arduino::MbedI2C::write(const uint8_t* data, int len) {
...
-	return -1;
+	core_util_critical_section_exit();
+	return rv;
 }
 
 int arduino::MbedI2C::available() {
-	return rxBuffer.available();
+	core_util_critical_section_enter();
+	int rv = rxBuffer.available();
+  core_util_critical_section_exit();
```
facchinm, line 139: Please fix the indentation here :slightly_smiling_face: 
Apart from this, everything looks good :+1:

facchinm (changes requested):

facchinm (approved):

PR #1044, opened 2025-03-27
feat: Opta pin names to match serigraphy

zfields:
Enables a user to intuit the pin name of the input pins, without having to reference this file.


manchoz:
LGTM

PR #1085, opened 2025-08-14
Fix Stella LEDs

marqdevx:
This fix, makes the white LED accessible.


pennam:
Note: Red Green and Blue LED are not mounted.

PR #1111, opened 2026-04-15
External se05x nicla

andreagilardoni:
This removes the mbedos component for se05x for nicla boards. In this way it will be possible to use an external common implementation for all cores


variants/NICLA_VISION/conf/custom_mbedtls_config.h
```diff
@@ -32,12 +32,6 @@
 
 /* clang-format off */
 
-#if defined(SSS_USE_FTR_FILE)
```
pennam, line 35: I don't think this change is needed, because the file is not compiled

variants/NICLA_VISION/defines.txt
pennam, line 1: this change goes together with libmbed.a rebuild

variants/NICLA_VISION/includes.txt
pennam, line 1: this change goes together with libmbed.a rebuild

variants/NICLA_VISION/mbed_config.h
pennam, line 1: this change goes together with libmbed.a rebuild

pennam (changes requested):

pennam (approved):

PR #1115, opened 2026-06-30
Arduino_H7_Video: stop reallocating rotate buffer on every redraw

spielp:
When the display is configured for rotation, lvgl_displayFlushing() reallocated rotated_buf to w*h*4 bytes on every flush, i.e. on every screen redraw. Because the dirty-area size varies frame to frame, this churned the heap continuously and could fragment it until the allocation failed mid-draw, crashing under sustained redrawing (e.g. on the Giga Display Shield). The result was also assigned straight back into rotated_buf, leaking the old block and dereferencing NULL on failure.

Grow the buffer only when a larger area arrives and converge to the max chunk size.


spielp:
How do I get the checks to rerun?

cc @per1234

per1234:
@spielp don't worry about it. The maintainers will be able to see that the compilation error was caused by a bug in the WiFiNINA library (https://github.com/arduino-libraries/WiFiNINA/commit/11157858b4914fbeaf3e0d9cb2a3a9925f947f51), not by the change you propose here.

PR #1118, opened 2026-07-05
fix: route PinName digitalWrite int values correctly

### LINKED ISSUE #1107: using digitalWrite() with a boolean instead of LOW/HIGH causes a affects a completely unrelated pin instead of the intended
## Summary
Calling digitalWrite() for PC_2 with a boolean true/false causes PA_8 to be toggled instead.
Calling digitalWrite() for PC_2 with a LOW/HIGH works as intended.

## Hardware
Board: Portenta H7
Core Version: Arduino-Mbed 4.4.1
Arduino IDE Version: 2.3.7
Arduino IDE CLI Version: 1.3.1

## Steps to reproduce
Attach an oscilloscope or logic analyzer to the following pins:
* PA_8 / D6 / PWM_0 
* PC2_C / A2 / ADC A2
* PC2 / A4 / ADC A4
* PC2 / D10 / SPI1_CIPO

and run the following code
```c
void setup() 
{
  pinMode(PC_2, OUTPUT);
}

void loop() {
  digitalWrite(PC_2, false);
  delay(1000);
  digitalWrite(PC_2, true);
  delay(1000);
}
```
My logic analyzer gives the following output:

Image

Compare this to the following functional code:
```c
void setup ()
{
    pinMode(PC_2, OUTPUT);
}

void loop ()
{
    digitalWrite(PC_2, LOW);
    delay(1000);
    digitalWrite(PC_2, HIGH);
    delay(1000);
}
```
Which gives the following output
Image

## Expected Behavior
Using true/false instead of LOW/HIGH should be equivalent.  
MOST IMPORTANTLY: doing so should absolutely not affect a different pin number on a different GPIO port altogether.

## Actual Behavior
Using true/false instead of LOW/HIGH on one port + pin has no effect on the intended port + pin but does have an effect on a different GPIO port + pin.

Old-Ding:
## Summary - add a `digitalWrite(PinName, int)` compatibility overload next to the existing `PinName` overloads - forward that overload to the existing `digitalWrite(PinName, PinStatus)` implementation  ## Why ArduinoCore-API provides a compatibility wrapper for `digitalWrite(pin_size_t, int)`. Without the matching `PinName` overload, calls like `digitalWrite(PC_2, false)` can convert the `PinName` value to an Arduino pin index path instead of using the mbed `PinName` path.  ## Validation - `git diff --check` - inspected ArduinoCore-API `Compat.h` and `Common.h` to confirm the `pin_size_t, int` compatibility overload and `PinStatus` definitions - inspected `wiring_digital.cpp` to confirm the existing `PinName, PinStatus` implementation is the correct target  I did not run a full Arduino board build locally because the required board/toolchain setup is not provisioned here.  Fixes #1107  ## CI notes
- The rerun of `compile-examples (arduino:mbed:nanorp2040connect, nanorp2040connect)` passed; all visible example-compilation matrix jobs are now green.
- `license/cla` is still pending.


Old-Ding:
CI investigation: the failing `nanorp2040connect` matrix job stops while compiling the unchanged WiFiNINA `SerialNINAPassthrough.ino` example, with `missing ')' in expression` at lines 50, 64, and 94. This PR only changes `cores/arduino/overloads.h` to add the `PinName` / `int` `digitalWrite` overload; no failing file or dependency path is touched by the diff.

I have not changed the overload for this unrelated example compilation failure. A maintainer rerun or the corresponding WiFiNINA / example-environment update is needed to clear the check.

